PR #2, opened 2022-02-05
repl + reify functions

felixroos:
- /repl contains simple website to play back and debug strudel patterns. can be built via `cd repl && npm run build` to /docs to make it available via github pages.
- reified functions in strudel.mjs for easier use

that piece of pastry is getting tasty :)


yaxu:
Thanks!

PR #7, opened 2022-02-07
fixed editor crash

felixroos:
fix: if the user entered an uncalled function like `pure("a5").slow`, the editor would crash. 
repl is built for "/".


yaxu:
Thanks!

PR #63, opened 2022-04-13
Dirtify plain values

felixroos:
This allows using existing "primitive value style" patterns with superdirt. Example:

```js
"<bd sd> hh".fast(2).superdirt()
```

or 

```js
"c3 e3 g3".s('superpiano').superdirt()
```

`.superdirt` will parse any primitive values to object format with "superdirt flavour" using `dirtify` and send them with `.osc`.
This way, `.osc` itself will just pump out values as is (only objectified). I also added `objectify` by default in .union, to make sure no error is thrown. The `dirtify` function transforms primitive values by those rules:

- numbers are always "n", with -60 offset (0 = c4 in tidal)
- note strings are transformed to numbers, also with -60 offset
- non note strings are treated as "s"
- objects will be treated as is, only the special key "value" will be dirtified itself. This allows objectified primitives to work as well

see util.test for tests

### DIFF
--- a/packages/core/strudel.mjs
+++ b/packages/core/strudel.mjs
@@ -1,5 +1,5 @@
 import Fraction from './fraction.mjs';
-import { isNote, toMidi, compose, removeUndefineds, flatten, id, listRange, curry } from './util.mjs';
+import { isNote, toMidi, compose, removeUndefineds, flatten, id, listRange, curry, objectify } from './util.mjs';
 
 class TimeSpan {
   constructor(begin, end) {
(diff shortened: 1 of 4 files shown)


yaxu:
Is this still current? Would be happy to help with it if I can ..

felixroos:
I just merged main into it.. would be nice to have this, although it's kind of breaking all existing .osc patterns for superdirt (changes .osc to .superdirt). also might require some testing if it still does what it's supposed to. If you want to have look please do!

yaxu:
Thanks, I'll have a play with it! `.superdirt` makes sense, to make room for a more generic OSC interface

yaxu:
Ok I've had a play with it and it all seems to work. I'm a bit unsure about it though, as it doesn't save many keystrokes and makes things more ambiguous, with potentially unexpected collisions between notes and sounds. I think it might be over-optimising on keystrokes.

felixroos:
yes i am unsure about it too, though i like the separation of superdirt / osc. allowing primitive values might be beginner friendly though.

felixroos:
closing this because it did not age well

Issue #46, opened 2022-04-12
Pseudo-randomness


yaxu:
A first step would be to implement the `rand` signal, then the other random functions can be implemented based on that.

Ideally this would give identical results to Tidal's RNG, although I guess depends on javascript and haskell having the same implementation of floating point numbers:
https://github.com/tidalcycles/Tidal/blob/main/src/Sound/Tidal/UI.hs#L46

Also in the process of writing this issue, I've noticed a couple of problems with tidal's `rand`:
https://github.com/tidalcycles/Tidal/issues/905

felixroos:
Good to know that randomness is only pseudo random, because it will make sure graphics that look further in to the future will have the same result

yaxu:
Yes there are some problems with this approach, like in this:
```javascript
s("bd*16").pan(rand).speed(rand)
```
High speeds would always be panned to the same place. But then you can get a different stream like:
```javascript
s("bd*16").pan(rand).speed(rand.slow(1.0001)
```

Issue #31, opened 2022-04-09
implement mainline tidal functions


felixroos:
There are many functions from tidal that have not landed yet. 

See this loooong list: https://github.com/tidalcycles/strudel/blob/main/featurelist.md#tidal-features-in-strudel

bpow:
`randcat` was implemented in 7fbd4527e

yaxu:
I think 'inhabit' is
```js
function inhabit(lookup, pat) {
	return pat.fmap(v => v in lookup ? lookup[v] : pure(silence))._squeezeJoin()
}
s(inhabit({a: "bd sd", b: "cp cp cp"}, "a b")).webdirt()
```

tedthetrumpet:
I know I asked about this already in another place and there is an answer of sorts, but: just to note that the Tidal function `run` is missing as well.

tedthetrumpet:
`slice` is missing, and not listed as missing.

yaxu:
Added `hurry` and `press`/`pressBy`

yaxu:
@tedthetrumpet looks like `run`, `slice` and `splice` are implemented now

tedthetrumpet:
Great, thanks!

jarmitage:
`smooth` -> #561 ?

https://tidalcycles.org/docs/reference/oscillators/#smooth

geikha:
note that `rolled`, `rolledBy` and `rolledWith` are missing

yaxu:
Pruned list ready for 1.0

felixroos:
pickF now here https://github.com/tidalcycles/strudel/pull/924 @yaxu just noticed after merging that pickF is listed under "Needs longer term work/thought"...

Issue #60, opened 2022-04-13
Standardise representation of values


yaxu:
Bring in unused `value.mjs`, and put all data in the value as an object, leaving context for things independent of value, like source code position.

felixroos:
related https://github.com/tidalcycles/strudel/pull/63

felixroos:
I'll try to summarize the problem with hap values.

- I'll add numbers to allow us to reference ideas quickly
- I'll add checkboxes to tell what is currently implemented

### 1. haps with object values

- [x] `n("0 1")` produces hap values `{ n: 0 }` and `{ n: 1 }`. 
- see [controls](https://github.com/tidalcycles/strudel/blob/main/packages/core/controls.mjs)

### 2. haps with primitive values

- [x] `"0 1"` will produce hap values `0` and `1`. 
- [ ] Alternatively, that syntax could also be represented with a reserved object key, producing hap values 
`{ value: 0 }` and `{ value: 1 }`, 

### 3. pattern methods that apply `union`

- [x] `n("0 1").room(.5)` produces hap values `{ n: 0, room: .5 }`
- see [controls](https://github.com/tidalcycles/strudel/blob/main/packages/core/controls.mjs)

### 4. pattern methods that assign to hap context

- [x] `"0 1".velocity(.5)` currently adds velocity value to `.context`, keeping values primitive.

- `scale`, set by `.scale`, referenced by `.scaleTranspose`
- `onTrigger` [called by REPL](https://github.com/tidalcycles/strudel/blob/f8739e341bc5babd2dcd8e608d1a2933979fe4e5/repl/src/useRepl.mjs#L51) when the event is scheduled, see https://github.com/tidalcycles/strudel/discussions/64 "How to implement a custom output". Currently referenced by all the outputs `.speak`, `.osc`, `.tone`, `.out`
- `instrument`, set by `.tone`, now unreferenced, can be removed
- `.type` set by `.tune`, `.xen` and `.tuning` to `frequency`, referenced by `getPlayableNoteValue`. Should be refactored to use `{ freq }` as value
- `velocity`, set by `.velocity`, referenced by `.midi`, `.tone`, `.adsr`, `.pianoroll`
- `locations`, set by `.withLocation`, `.withMiniLocation` to indicate source relevant source code locations, referenced by `markEvent``
- `color`, set by `.color`, referenced by `.pianoroll`, `markEvent`

### 5. primitive haps = (2) + union method (3)

[...]

PR #69, opened 2022-04-15
First effort at rand()

yaxu:
Seems to work, but 
```
"c4*16".velocity(rand)
```
only gives one random value per cycle. This works as expected:
```
"c4*16".velocity(rand.struct("t*16"))
```
This is probably an issue with applicative (or maybe monad) or the way that velocity is combined with value.

Contributes to #46 and #31

### DIFF
--- a/packages/core/signal.mjs
+++ b/packages/core/signal.mjs
@@ -28,3 +28,34 @@ export const square2 = square._toBipolar();
 
 export const tri = fastcat(isaw, saw);
 export const tri2 = fastcat(isaw2, saw2);
+
+
+// random signals
+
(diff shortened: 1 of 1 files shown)


yaxu:
Applicative looks fine:
```
console.log("3 4 2".fmap(a => b => a + b).appLeft(rand).firstCycle())
0: Object { whole: {…}, value: 2.168623458594084, stateful: false, … }
​1: Object { whole: {…}, value: 4.2604806162416935, stateful: false, … }
​2: Object { whole: {…}, value: 1.9786392766982317, stateful: false, … }
​```
So I think this will be solved with #60

Issue #53, opened 2022-04-12
binary pattern reset function


felixroos:
I don't know if this exists in tidal, but it would be fabulous if there was something like that:

```js
"c3 e3".reset("<x [x x]>")
```

being equivalent to `"<[c3 e3] [c3 c3]>"`. So the binary pattern will reset the input pattern each time there is a true event. 

In the example, this is looks pretty boring, but for composing, this opens many doors, for example making sure the bass lines always lands on the root when the chord / scale changes. 

Here is an [example](https://tinyurl.id/lzhOu) of a bass line that doesn't land on the 1 for the last note:

```js
"0 4".scale(
  sequence('C major', ['D dorian', 'G mixolydian']).slow(2)
)
```

The output is equivalent to `"<[c3 g3] [d3 d4]>"` with the indices `"<[0 4] [0 4]>"`.

Ideally, the last note should be a g3, index 0 in 'G mixolydian' which starts there.
A human bass player would almost always play the root note on the first beat of a new scale / chord.

To mimic that behavior, we could use the not yet existing reset function:

```js
const scales = sequence('C major', ['D dorian', 'G mixolydian']).slow(2)
"0 4".reset(scales).scale(scales)
```

This will reset the "0 4" each time the scale changes (any scale is true), creating this index pattern

```js
"<[c3 g3] [d3 g3]>"
// "<[0 4] [0 0]>"
```

... which is what we want in that case. This is not only useful for basslines, but also arpeggios or glitchy retriggering patterns.
This will be fantastic for any type of dynamic groove!

yaxu:
No this isn't in tidal but has been requested before.

Instead of resetting the clock to `0`, you might want to reset it to the next bar/quantum/modulo of the cycle count.

I think this is related to patterning the `cps`, as this resetting is a kind of clock operation. Often you'd want to do this to only part of a pattern. But that might be true of `cps`as well, i.e. you might want to change the `slowing` of a part of pattern without skipping events.

yaxu:
Solved in https://github.com/tidalcycles/strudel/pull/88/

PR #88, opened 2022-04-22
Reset, Restart and other composers

felixroos:
https://github.com/tidalcycles/strudel/issues/53


yaxu:
I think reconfiguring it into a join would be good, like in [this example](https://is.gd/rsTD1Z). Then we can easily make `_opReset` and `_opRestart` and add to the pattern composer family https://github.com/tidalcycles/strudel/blob/main/packages/core/pattern.mjs#L751-L767

yaxu:
Now we have:

 . | In | Out | Mix | Squeeze | SqueezeOut | Trig | TrigZero |
| - | --- | -- | --- | --- | --- | --- | --- |
| add | add/addIn | addOut | addMix | addSqueeze | addSqueezeOut | addTrig | addTrigZero | 
| sub | add/subIn | subOut | subMix | subSqueeze | subSqueezeOut | subTrig | subTrigZero |

and the same for these operators:
```
set keep keepif mul  div  mod  pow  lt  gt  lte  gte  eq  eqt  ne  net  and  or _and  _or  _xor  _lshift  _rshift func
```
`keepif` only keeps for values on the 'outside' pattern (on the right) that are true (truthy), otherwise the value is undefined (for later removal with `_removeundefined()`).

`lt` is `<`, `gte` is `>=`, `eqt` is `===` etc.

All operators 'op' default to `opIn`, i.e. applying operator into the pattern on the left.

Then there are these aliases to give functions that deal only with structure not particular values:
struct = 
| alias | for |  
| - | - | 
| struct | keepifOut |
| structAll | keepif |
| mask | keepifIn |
| maskAll | keepifOut |
| reset | keepifTrig |
| resetAll | keepTrig |
| restart | keepifTrigZero |
| restartAll | keepTrigZero |

The `All` variants use all events from the outside pattern, not just the 'true' ones.

I'm happy with all this now. I'm just unsure whether `struct` etc should default to use all events, with the variant called e.g. `structBinary`

felixroos:
I like the `in` `out` wording! Also good to have the aliases you mentioned. I guess I'd need to discover the utility of all those functions by using them. Maybe we will find some more aliases later.

yaxu:
I think there will be some naming tweaks as well later, as we try them out.. But I think this is a good time to merge?

felixroos:
yes, let's merge

PR #97, opened 2022-04-25
Codemirror 6

felixroos:
Started moving to Codemirror 6. 

Before merging, highlighting must be fixed for overlapping events: Currently, when the same source location has overlapping events, the highlighting will flicker and probably switch off too early


felixroos:
Added some fixes + missing features from the old notation:

- moved highlighting logic to useHighlighting + added cm6 to mini repl
- added support for custom event colors
- fixed a bug where codemirror would try to highlight chars that have been deleted, which would cause a crash
- removed old codemirror packages

The new logic behaves a little bit differently in some cases, which might be related with a bug in the querying for small chunks of time. I'd say we can merge when that's resolved

PR #101, opened 2022-04-27
Change to Affero GPL

yaxu:
After discussion, @felixroos and I would like to switch from [GPL](https://www.gnu.org/licenses/gpl-3.0.en.html) to the [Affero GPL](https://www.gnu.org/licenses/agpl-3.0.en.html) for it's coverage of web services. This will also match the license for [Hydra](https://github.com/hydra-synth/hydra). It's best to do this now before there are more substantial contributions from others.

Tagging people with merged/ongoing PRs: @puria @bwagner @ChiakiUehira @kindohm @charlieroberts @mindofmatthew @filippoguida

Please shout if you object to your changes being distributed under AGPL rather than GPL, thanks!


yaxu:
I added the AGPL headers with this script.

Please take a look what you think (it'll be harder to make formatting tweaks etc later).

```perl
#!/usr/bin/perl -w

use strict;
use File::Basename;

my @fns = `find packages/ repl/ -regex '.*\.m?js'|grep -v node_modules`;

undef $/;
my $licence = <DATA>;

foreach my $fn (@fns) {
  chop($fn);
  my $base = basename($fn);
  open FILE, "<$fn";
  my $source = <FILE>;
  close(FILE);
  open FILE, ">$fn";
  my $tmp = $licence;
  $tmp =~ s/replaceme/$base/sg;
  print FILE $tmp;
  print FILE $source;
  close FILE;
  unlink("$fn.tmp")
}

__DATA__
/*
replaceme - <short description TODO>
Copyright (C) 2022 <author(s) TODO> and contributors
This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more details. You should have received a copy of the GNU Affero General Public License along with this program.  If not, see <https://www.gnu.org/licenses/>.
*/
```

Issue #28, opened 2022-04-09
Codemirror 6


felixroos:
It would be nice to update the REPL to Codemirror 6.
Here are some examples of what it brings to the table: https://codemirror.net/6/examples/
I especially like:

- [decoration widgets for custom ui elements](https://codemirror.net/6/examples/decoration/#boolean-toggle-widgets)
- [collab editing](https://codemirror.net/6/examples/collab/)
- [custom gutters](https://codemirror.net/6/examples/gutter/)

I already tested [react-codemirror6](https://www.npmjs.com/package/react-codemirror6 ) and it's pretty easy to integrate.
You can see it in action as part of [this post](https://loophole-letters.vercel.app/strudel) (The highlighting is still buggy when an event is triggered while already highlighted.).

felixroos:
this should be a good starting point: https://github.com/felixroos/felixroos.github.io/blob/main/components/strudel/CodeMirror.jsx

yaxu:
A big plus is that this will make strudel more accessible e.g. to people using screen readers. It'd be good to do some testing around that.

PR #102, opened 2022-04-28
Work on Codemirror 6 highlighting

mindofmatthew:
Here's an implementation of highlight marking that queries the pattern every frame. There are certainly some performance optimizations that could be done (plus some nuances of exactly how it should behave when the user starts/stops the pattern), but I think this is a good conceptual starting-point and an implementation that seems to perform pretty well.

Let me know if you've got any thoughts!

### DIFF
--- a/repl/src/App.js
+++ b/repl/src/App.js
@@ -1,5 +1,5 @@
 import React, { useCallback, useLayoutEffect, useRef, useState, useEffect } from 'react';
-import CodeMirror6, { highlightEvent } from './CodeMirror6';
+import CodeMirror6, { setHighlights } from './CodeMirror6';
 import cx from './cx';
 import logo from './logo.svg';
 import playStatic from './static.mjs';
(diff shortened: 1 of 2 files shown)


repl/src/App.js
```diff
@@ -118,6 +119,29 @@ function App() {
     return () => window.removeEventListener('keydown', handleKeyPress);
   }, [pattern, code, activateCode, cycle]);
 
+  useEffect(() => {
```
felixroos, line 122: looks good! we could also move this to a custom hook like `useHighlights({pattern,view,started})` or similar, then we could reuse it in the mini repl

repl/src/CodeMirror6.jsx
```diff
@@ -19,60 +18,29 @@ const highlightField = StateField.define({
...
-          highlights = highlights.update({
-            filter: (f, t, value) => {
-              if (f === e.value.from && t === e.value.to) {
-                return false;
-              }
-              return true;
-              // console.log('filter', f,t,value, e.value.from, e.value.to);
-            },
-          });
+        if (e.is(setHighlights)) {
```
felixroos, line 22: that looks much simpler than what I fabricated :D didn't know you could just set the Decorations directly

repl/src/CodeMirror6.jsx
```diff
@@ -19,60 +18,29 @@ const highlightField = StateField.define({
...
+          highlights = Decoration.set(
+            e.value
+              .flatMap((event) => event.context.locations || [])
+              .map(({ start, end }) => {
+                let from = tr.newDoc.line(start.line).from + start.column;
```
felixroos, line 27: This should make `offsetToPosition`/ `positionToOffset` obsolete :)

felixroos (approved):
thank you!

felixroos:
related https://github.com/tidalcycles/strudel/issues/28

Issue #34, opened 2022-04-10
block based evaluation


felixroos:
Think about if we should implement block based evaluation, like in mainline tidal.
Currently, the whole code is reevaluated when hitting ctrl+enter.
To make that work, some way of naming patterns needs to be found. One idea is using the rarely used javascript label statements:

```js
d1: "bd sn"

d1: "bd ~"
```

If we run the fist line, and then the second, only the second should play.

To make that work, the runtime needs to store patterns in some map that stores `{ [name]: pattern }`.
When a line is evaluated, the pattern map is updated.
 After each evaluation all patterns in the map are stacked together and queried.

One question is when non pattern code should be evaluated? Example:

```js
const p = (await piano()).toDestination();

d1: "c3 e3 g3".tone(p)
```

Here, `d1` references `p`, so we need to have that evaluated too. 
It would make usage much more complicated if the user is expected to run the first line before the second otherwise it crashes.

felixroos:
somewhat related: https://extemporelang.github.io/docs/overview/using-extempore/

felixroos:
before it gets forgotten, from chat:

yaxu:
With the usual tidal editor plugin approach, blocks are pasted into a repl. it makes it more of a conversation but it's difficult to keep aware of which code is being run
with the strudel approach it's just the final line that brings things together. If there's mininotation in play it's pretty obvious which bits are running but it's a bit awkward having to scroll to this line at the bottom being in charge of everything
The feedforward approach is more of a structural approach, where blocks automatically get tagged with numbers and can be switched on and off with a shortcut. You also get a VU meter per block. You can only write patterns though, not e.g. definitions of a new function to use
With tidal editor plugins you have to do the work of assigning numbers to patterns and it's really easy to accidentally replace a running pattern that you didn't want to. Unexpected breakdowns can be nice, but..
Another approach would be tiling multiple editors and having one editor per tile. Siren did something like this but mixed with a old school tracker paradigm https://toplap.org/2017/04/12/siren/
It could be interesting to have a semi-structured approach like if you did ctrl-enter on s("bd sd") it would be transformed into s("bd sd").send(3) , where '3' is a unique number. Then you could toggle it with ctrl-3 and it'd toggle between that and  s("bd sd").mute(3).
this could be called a 'performance mode'
the send(3) would do the work of adding the pattern to be scheduled

froos:
[...]

PR #110, opened 2022-05-05
add createParam + createParams

felixroos:
as requested by @guillaume-leo, this PR contains the methods `createParam` + `createParams` to be able to use custom params, e.g. for osc:

```js
const dudu = createParam('dudu'); // for single params
const { dada, dodo} = createParams('dada', 'dodo'); // multiple params
```

the param can then be used either to start a pattern, or as a pattern method:

```js
dudu('xxx').dada('yyy').osc(); // {dudu:'hehe', dada: 'yyy'}, osc message ['dudu','hehe','dada','yyy']
dodo('xxx').dudu('yyy').osc(); // {dodo:'hehe', dudu: 'yyy'}, osc message ['dodo','hehe','dudu','yyy']
```

the param is also patternified by default:

```js
dudu("a [b c]") // {dudu:'a'}, {dudu:'b'}, {dudu:'c'}, 
```

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -291,4 +291,13 @@ generic_params.forEach(([type, name, description]) => {
   Pattern.prototype[name] = _setter(controls[name]);
 });
 
+// create custom param
+controls.createParam = (name) => {
+  Pattern.prototype[name] = _setter(controls[name]);
+  return (...pats) => _name(name, ...pats);
+};
+
+controls.createParams = (...names) =>
+  names.reduce((acc, name) => Object.assign(acc, { [name]: createParam(name) }), {});
+
 export default controls;


yaxu:
I think this could be useful enough to name them `param`/`params`

Also if `param` is called with > 1 parameter, how about it treats the remaining parameters as a sequence. Allowing:

```js
var dudu = param('dudu'); 
```
but also:
```js
param('dudu', 3, 4, 2).speed(2) 
param('dudu', "3 4 2").speed(2) 
```
Also as a pattern method:
```js
note("a3 a4").param('dudu', "3 4 2")
```
As sometimes you might want to send a one-time-only parameter.

In fact in tidal, this function is just `p` (although they have to be specialised with a type so `pF` for floats,, `pS` for strings etc)

felixroos:
In that case `param('dud')` shouldn't add `.dudu` to Pattern i suppose?

yaxu:
Hmm.  Maybe `param('dudu')` should but `param('dudu', <pattern>)` shouldn't?

felixroos:
or just separate `param` and `createParam(s)` ?

yaxu:
Yes could be..

felixroos:
or maybe it looks cooler to do `Pattern.createParam('dududu')` . I think most of the time, there will be a hand full of custom patterns used that are created beforehand as a setup step, but still the `param(name, ...pats)` would be useful for inline usage and or patternified param names

yaxu:
Hmm I think this is tangentially related to #34 . Being able to stick `pF` in patterns is particularly useful in tidal due to block based evaluation. When you evaluate everything then there's less cost to having separate statements.

yaxu:
(I talked in a circle here but happy for you to merge whatever feels right!)

felixroos:
I'll merge this now, we can add the inline param function in a separate feature

PR #121, opened 2022-05-28
Webdirt

felixroos:
adding support for WebDirt !  

(based on in-source-doc branch)


felixroos:
TODO: add support for clean-samples / estuary style resources.
Probably need to handle sample loading on strudel side and pass buffers to webdirt.
This is how buffers can be loaded: https://codesandbox.io/s/sample-playback-04o520?file=/src/index.js

Issue #115, opened 2022-05-16
pianoroll() options


jarmitage:
For example:

`.pianoroll(axis, offset, direction)` where:
- `axis` is e.g. 0 or 1 corresponding to horizontal and vertical orientation (default 0)
- `offset` a float from 0-1 will move `now` along some % of the selected axis (default 0.5?)
- `direction` will change the scroll direction from left/right in horizontal and up/down in vertical orientations

This is basically inspired by Hydra. So another approach altogether would be to send `pianoroll()` into Hydra, and then the above wouldn't be necessary.

felixroos:
pianoroll already supports some option in object format. the default options are

```js
  timeframe = 10, // 10s lookahead
  inactive = '#C9E597', 
  active = '#FFCA28',
  background = '#2A3236',
  maxMidi = 90,
  minMidi = 0,
```

The options you are proposing look good, and they should be not to difficult to implement with the current draw logic.
I think it would be easiest to pull them into the object, and not use arguments.

PR #131, opened 2022-06-05
Pianoroll enhancements

felixroos:
I added the following options to `.pianoroll`:

```js
{
cycles = 4, // number of cycles to show
playhead = 0.5, // position of playhead, 0.5 = center
vertical = 0, // if 1, time will be on the y axis
autorange = 0, // if 1, the range of values will be deduced from the active cycles (instead of minMidi / maxMidi)
fold = 0, // if 1,  there will be no gaps between values (like ableton fold), autorange will be ignored
overscan = 1, // how many extra cycles will be queried outside the viewport, can be used to get a more stable autorange without needing to set cycles higher
flipTime = 0, // if 1, time will flow in the opposite direction
flipValues = 0, // if 1, value range will be flipped
hideNegative = 0, // if 1, haps with negative events will be hidden at start
timeframe, // now deprecated (but still works)
}
```

### DIFF
--- a/packages/tone/draw.mjs
+++ b/packages/tone/draw.mjs
@@ -20,7 +20,7 @@ export const getDrawContext = (id = 'test-canvas') => {
   return canvas.getContext('2d');
 };
 
-Pattern.prototype.draw = function (callback, cycleSpan, lookaheadCycles = 1) {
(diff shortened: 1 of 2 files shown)


felixroos:
fixes #115

Issue #111, opened 2022-05-08
Fix legato and duration


yaxu:
The current `legato` implementation manipulates haps in a way that can lose data.

A fix would be to copy tidal's behaviour - store legato in a normal value key and then deal with it in the scheduler.

felixroos:
Isn‘t this problem of loosing data also related to all other methods that use withHapTime? Iirc the problem was there is no way of knowing if an event part ever had an onset or not. But maybe there was another problem I don’t remember anymore. Generally, I like the possibility of controlling the hap length directly, because we then don’t need to react to that param for every output separately

yaxu:
It's not a problem with others that I know of, because where `withHapTime` is expanded, `withQueryTime` is reduced by the same factor. The particular problem with this approach to legato is that we don't know by how much the query time should be expanded by.

We can still solve this once for everything though, as a simple pre-processing step that changes the duration of events before calling the callback. The only problem is that superdirt will do this a second time if it sees the `legato` parameter. So we need to remove it or reset it to `1` when the durations are changed.

felixroos:
What are the consequences of not reducing withQueryTime by the same factor withHapTime is expanded? Can you think of an example problem this might cause? The thing with the pre processing step sounds like a good idea

yaxu:
The consequence is events appearing or not depending on where you query:

A part from 0 > 2:
```js
slowcat(1,silence).legato(2).queryArc(0,1)
```

Disappeared:
```js
slowcat(1,silence).legato(2).queryArc(1,2)
```

This will result in glitches and inconsistent results, especially when it comes to combining it with other patterns.

yaxu:
The `duration` method has the same problem at the moment. This is kind of possible to fix for `duration`, as for query timespan (a,b), you could query (a-duration,b), to be sure of finding events that are expanded into the query timespan.. That would mess with the values returned from continuous events though.

yaxu:
I've fixed duration in https://github.com/tidalcycles/strudel/commit/961a80dfe7ec912da9e365b2dc9eff2d1ba84d1f, but this breaks the 'jemblung' tune, as it has a pattern of the form `"12".duration(0.05).add(12)`, that doesn't work when duration is promoted from metadata to data.

Perhaps it would be best to change all examples to use explicit parameter names for patterns, to make things clearer for beginners.

yaxu:
[...]

PR #148, opened 2022-06-27
Sampler optimizations and more

felixroos:
This PR includes a multitude of fixes and features. The main ones are

### add `note` to repitch samples + `choke` to cut samples to hap duration:

```js
samples({ 
  flbass: '00_c2_finger_long_neck.wav' 
}, 'github:cleary/samples-flbass/main/')

note("<8(3,8) <7 7*2> [4 5@3] 8>".sub(1).scale('A1 minor'))
.s("flbass").choke(1)
.out()
```

- `.out` will repitch the sample relative to `c3`. 
- choke makes sure the samples won't overlap. this is different from cut, as it will also respect rests

### `samples` now also supports object format for pitch declaration

```js
samples({ 
  piano: {
    A3: 'A3v8.mp3',
    C4: 'C4v8.mp3',
    A4: 'A4v8.mp3',
    C5: 'C5v8.mp3',
  }
}, './piano/')

note("<8(3,8) <7 7*2> [4 5@3] 8>".sub(1).scale('A3 minor'))
.s("piano").choke(1)
.out()
```

- `.out` will now find the nearest sample for the given pitch and repitch relative to that pitch
- to simplify note declaration, now added support to notate sharps with "s" as an alternative to "#"

## Prebake Script with Piano

- added piano samples to replace tone piano, using same samples (Creative Commons, see https://archive.org/details/SalamanderGrandPianoV3)
- added prebake script to REPL, to load default sounds, where there is now `.piano`  to load the piano samples with choke and stereo effect, as an alternative to tonejs piano
- refactored tunes to always use the new .piano instead of tonejs piano

## Version Safety

After having problems with version collisions:

- log that @strudel.cycles/core is loaded, warn if it's loaded multiple times
- warn when a Pattern fails instanceof check

## Other

- can now use ":" in mini notation to set "n" for "s" or "note"

## Migrating away from Tone.js

With the above features, many more tunes can now be rewritten without Tone.js, which I will do in the future.
When all are refactored, we could think about changing what happens when no output is chosen.
Currently, the default behavior is playing the pattern with the default synth, supporting only primitive hap values.
It might make more sense to use `.out` by default in the future, or just log the events to the console, like tidal.


yaxu:
Ace!! How about using `legato` instead of `choke`?

felixroos:
`legato` has no effect on samples, as samples will always play until the sound buffer ends, which can exceed the `hap.duration`, creating overlaps. `legato` will only modify the `hap.duration`, relative to itself, so `legato(2)` will multiply the duration by 2, but the sample still won't care about that. Setting `choke(1)` (`1 === true`) will tell the sampler: please care about the `hap.duration` and ignore the buffer duration. Having samples always play till the end can still be useful because overlaps might be wanted in the case of drums.

yaxu:
I think `legato` should 'just' be another parameter in the value object rather than metadata, then the synthesiser can decide how best to interpret it. Then it can work the same way for samples and synths, and solve #111. This approach works well with tidal/superdirt.
Btw superdirt applies legato slightly differently for samples and synths - for samples legato is default 'off' (i.e. the full sample plays), and for synths it's default `1`.

yaxu:
Re-reading your comment, my proposal is that as with superdirt, the default for samples should be as you describe - play the whole sample. When legato is set, it can 'choke' the sample relative to hap duration. But the hap duration shouldn't be changed until it is triggered. That is, legato doesn't change the duration of the hap, but the way it is 'played'.
Whether legato is represented in a visualisation (e.g. pianoroll) depends on whether you want to visualise the underlying pattern structure or the resulting sounds. I guess that should be a pianoroll option.

yaxu:
Yes that sounds like a good path to me.

I realise it's a term from percussion but I'm not super keen on the word `choke` though, and as a concept cymbal choking is closer to how `cut` works. How about `clip`?

felixroos:
I chose `choke` because that's how it was called on my good old MPC (there were choke groups, similar to cut groups in superdirt). But no strong feelings here, I am also fine with `clip`. Then I'll rename it

yaxu:
Thanks. I do find it a bit funny how many different functions we have in tidal/strudel with names that mean 'cut things into pieces', good to add another one ;)

felixroos:
Let's keep this one secret until it's away again :)

PR #150, opened 2022-07-11
add webdirt drum samples to prebake for general availability

larkob:
This loads all default webdirt samples via prebake, so you can play a few drum samples with `out` as well as `webdirt`. This is mostly to make Strudel more beginner-friendly, but also so you can easily apply webaudio methods like `velocity` that webdirt doesn't support yet.

### DIFF
--- a/repl/package.json
+++ b/repl/package.json
@@ -7,7 +7,7 @@
     "start": "vite",
     "build": "vite build",
     "preview": "vite preview",
-    "test": "mocha ./src/test --colors",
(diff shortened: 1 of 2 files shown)


larkob:
I'm using the experimental fetch feature (supported in node 16 LTS and 18), so there's no need to keep track of changes to the  `EmuSP12.json` file that's coming from the webdirt repo. However, this will break possible compatibility with earlier versions of node.

PR #152, opened 2022-07-12
bpm and flatcat functions

larkob:
I've added two functions that are particularly useful for beginners:

* `bpm` allows to set the tempo in BPM instead of CPM. The underlying idea is that a standard cycle has 4 elements or beats (corresponding to a 4/4 time signature).
* `flatcat` is like `cat` but allows to concatenate patterns without worrying whether the items are arrays or not. It applies flattenDeep to all items before concatenating. I use it to lay out a sequence of bars in a more familiar song structure fashion. I could certainly use `cat` with the `...` spread operator, but then I would always need to know which item is an array and which is not (or test for it, which just convolutes the code).

```
var intro     = ["~ x x x", "~ x ~ x"]
var verse   = ["x ~ ~ ~", "x ~ ~ ~"]
var chorus = [intro[0],"x x x x"]
var bridge  = "x ~ x ~"
s("hh").n(1).struct( flatcat(intro, verse, chorus, bridge, verse, chorus, "x ~ ~ x") ).bpm(120).webdirt()
// the same with cat:
//s("hh").n(1).struct( cat(...intro, ...verse, ...chorus, bridge, ...verse, ...chorus, "x ~ ~ x") ).bpm(120).webdirt()
```

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -10,7 +10,7 @@ import Hap from './hap.mjs';
 import State from './state.mjs';
 import { unionWithObj } from './value.mjs';
 
-import { isNote, toMidi, compose, removeUndefineds, flatten, id, listRange, curry, mod } from './util.mjs';
+import { isNote, toMidi, compose, removeUndefineds, flatten, id, listRange, curry, mod, flattenDeep } from './util.mjs';
 import drawLine from './drawLine.mjs';
 
 /** @class Class representing a pattern. */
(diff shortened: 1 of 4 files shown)


felixroos:
I would also avoid the default assumption of 4 beats per cycle. How about a second param that specifies that, like: `.bpm(120, 4)`?
this would be equivalent to `.cpm(120).slow(4)`.
For pattern composition, it might be a good idea to look at https://tidalcycles.org/docs/reference/composition/#ur
Is there an equivalent of `flatcat` in tidal? (btw there is also `Array.flat` which does what `flattenDeep` does)

yaxu:
Yes I wondered about passing beat per cycle to bpm too. It seems though that strudel is growing a kind of convention where methods only take one parameter, so that pattern sequences can be passed. 

Longer term I think cps probably shouldn't be defined in terms of `fast`/`slow`, but in terms of changing tempo state in the scheduler as with Tidal. Otherwise changing tempo with `cps` jumps around the pattern timeline. (Perhaps there could also be stateful equivalents of fast/slow)

larkob:
I'm closing this PR now because it isn't ready to be merged and I can just create my own helper functions for now (using the `gist` function).

PR #162, opened 2022-07-24
update to tutorial documentation

larkob:
Added Euclidian documentation, removed non-existent append function.


felixroos:
Thanks :)

PR #165, opened 2022-07-29
incorporate elements of randomness to the mini notation

bpow:
*Work in progress*, not ready for merge

Currently implements the `?` operator (degradeBy). Trying to implement the random choice notation (pipe-separated elements in a list), but needs `randcat` in order for that to work. I think I have a working peggy parser implementation, though.


bpow:
I'm going to leave this titled as (WIP) for now, but these commits do result in implementation of the `?` notation for degradeBy and `|` notation for random choice among cycles.

I've rebased against main for some clarity and deferred updating of `krill-parser.js` to its own commit rather than in each of the first 2 commits to avoid extra churn to this generated file. Happy to just squash everything together if you'd prefer.

Rather than just use `randcat`, I followed the logic of [5501a8 from Tidal](https://github.com/tidalcycles/Tidal/commit/5501a8df7a89399076283d8d401b13698ad2bc5c) to keep a "seed" as part of the parser state and move within `rand` by a fraction of that seed (incremented each time it is used). This prevents things like having 'c4? e4? g4?' having each note use the same segment from `rand` (so they would either all be played or all muted in a given cycle). Similarly, it keeps `"[c3|e3|a3],[c4|e4|a4]"` from just playing octaves each time -- instead it sometimes plays octaves, but randomly sometimes plays pairs of notes from the A minor triad, one each in octave 3 and 4.

I'm not sure exactly how to best write tests for this since there's randomness involved. Doing so might have to involve either testing that one or another of the alternatives is chosen as the first cycle, or injecting a specific `rand` (with specific time seed) into the parser.

yaxu:
Great! One way to test is to add an example that uses `?`, then a test is automatically done to check that the values are always the same.

bpow:
I'm ready to have someone else look over this. It is all rebased to main as of now, has some tests within packages/mini and also a snapshot-based test.

There's probably some more work that could be done to the pegjs grammar (this doesn't address #176, for instance, but I think that involves enough other things to warrant a separate pull request flow).

I tried to have the commits make logical sense, but would be happy to squash things if that's the local convention.

felixroos:
This looks really good! no objections here. We don't have any conventions when it comes to squashing commits, I think it's pretty readable as it is. Thank you so much!

felixroos:
I just noticed some strange behavior in the bossaRandom example:

```js
const chords = "<Am7 Am7 Dm7 E7>"
const roots = chords.rootNotes(2)

stack(
  chords.voicings(['F4', 'A5']).struct(
  ` x@2   ~ x ~ ~ ~ x |
    x?  ~ ~ x@3   ~ x |
    x?  ~ ~ x ~ x@3`).note().gain(.5),
  roots.struct("x [~ x?0.2] x [~ x?] | x!4 | x@2 ~ ~ ~ x x x")
  .transpose("0 7").note()
).slow(2).pianoroll().s('sawtooth').out();
```

With this snippet, I consistently get a missed event on the downbeat of the second playthrough of the chords (when starting from a fresh load). The odd thing is that the downbeat of that first phrase has no "?" so it should play in every case. The problem seems to vanish when I remove the chords

Issue #168, opened 2022-08-02
Can't use import in REPL


debrisapron:
I want to import code from e.g. jsdelivr in the REPL but I get an `Unexpected token "import"` message. Does the JS parser the REPL uses not support import? This would be super helpful for prototyping plugins for Strudel without the whole ceremony of forking the repo, creating a dev environment etc etc.

felixroos:
imports are not supported (yet?). We use shift-ast, which supports import, but for that we need a custom implementation of loading scripts on the fly, which we do not have yet. 

There was a discussion recently about loading scripts: https://github.com/tidalcycles/strudel/discussions/153 (search `loadScript`) Here is an example of loadScript: https://strudel.tidalcycles.org/?TO3z8wnvE2p3 it's just an eval so no full blown module importer. I haven't yet dug into how a module loader could work client side (yet), any help / tips are welcome

In case you want to do the ceremony, here is a setup guide: https://github.com/tidalcycles/strudel/blob/main/CONTRIBUTING.md#project-setup

debrisapron:
I've actually done basically this exact work in another REPL-type project, would you want a PR?

felixroos:
yes please! which types of modules did you manage to load with it?

debrisapron:
Any valid native JS module works fine. Like you could do something like `import * as strudel from "https://cdn.jsdelivr.net/gh/tidalcycles/strudel/main/packages/core/index.mjs"` and you should be good to go

felixroos:
that would be nice! feel free to ask any questions that may arise

debrisapron:
OK so after further investigation this is not really possible with the current REPL. At the moment, the `shapeshifter.mjs` module in @strudel/eval makes a lot of mutations to the AST which, while I'm sure it makes for a fast & sexy live-coding experience, breaks a lot of javascript features. For example, currently all string literals are run through `mini`, which obviously breaks any kind of import function, because URLs are strings.

Off the top of my head, here are a few possible solutions:

1. Get rid of the auto`mini`fication magic completely and just use a tagged template literal, e.g. something like:
```
m`[B3@2 D4] [A3@2 [G3 A3]] [B3@2 D4] [A3]`.transpose(1).whatever().blah()
```
This is still pretty terse & doesn't break JS semantics.

[...]

PR #171, opened 2022-08-04
Amend shapeshifter to allow use of dynamic import

debrisapron:
Whilst implementing the standard import syntax (`import foo from "./bar"`) is problematic due to the statically analyzed nature of these imports, the [dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import) syntax works fine in this context. Unfortunately the shift parser does not support dynamic import and actually throws an error when it sees the keyword, so I have implemented a mildly hacky solution which swaps it out before parsing, then swaps it back in again afterwards.

You can see it working by running this code in the REPL:

```
const { default: confetti } = await import('https://cdn.skypack.dev/canvas-confetti');
confetti();

stack(
  "<C^7 F^7 ~> <Dm7 G7 A7 ~>"
   .every(2, fast(2))
   .voicings(),
  "<c2 f2 g2> <d2 g2 a2 e2>"
).transpose("<0 2 3 4>")
```

### DIFF
--- a/packages/eval/shapeshifter.mjs
+++ b/packages/eval/shapeshifter.mjs
@@ -34,7 +34,7 @@ export const wrappedAsync = true;
 
 export default (_code) => {
   const { code, addReturn } = wrapAsync(_code);
-  const ast = parseScriptWithLocation(code);
+  const ast = parseScriptWithLocation(disguiseImports(code));
   const artificialNodes = [];
   const parents = [];
   const shifted = replace(ast.tree, {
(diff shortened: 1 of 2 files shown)


felixroos:
Great! 
While this looks a bit hacky, I don't see a problem it might cause. Just did a bit of research and found out that [shift](https://shift-ast.org/) only supports ECMAScript up to ES2019, but dynamic import was added in ES2020. So this is currently the only solution with shift. Sadly, shift seems [a bit unmaintained atm](https://github.com/shapesecurity/shift-parser-js/issues/452#issuecomment-1101739466).. 

Maybe we could move to another parser / transformer lib in the future, like [babel standalone](https://babeljs.io/docs/en/babel-standalone) (which is based on acorn, which supports ES2020) or if that's too heavy just the packages we need (@babel/parser @babel/traverse @bable/type @babel/generate + any plugin packages we need, like [top level await](https://babeljs.io/docs/en/babel-plugin-syntax-top-level-await).

But for now, this PR looks good and is a valuable addition, so thank you :) should we merge it then?

debrisapron:
Let's do it! But yes, I think the permanent solution is to switch to @babel/parser, it's strongly optimized for this kind of quick and simple AST manipulation, extremely battle-tested & well documented + you can even support sexy experimental JS syntax if you like (e.g. https://github.com/tc39/proposal-pipeline-operator)

felixroos:
fixes #168

Issue #143, opened 2022-06-24
consider using @uiw/react-codemirror


felixroos:
https://github.com/uiwjs/react-codemirror seems to be maintained more actively, compared to what we're currently using (https://www.npmjs.com/package/react-codemirror6). it also has some extra themes. switching should not be difficult

Issue #125, opened 2022-06-01
cannot evaluate after undo


felixroos:
codemirror6 does some thing that won't let evaluating after undo:

1. enter some valid code
2. evaluate
3. make a change
4. evaluate
5. undo
6. evaluate => won't work

this is pretty irritating

felixroos:
this seems to be an issue of react-codemirror6 or codemirror see https://github.com/tbjgolden/react-codemirror6/issues/9

felixroos:
fixed with https://github.com/tidalcycles/strudel/pull/173

PR #173, opened 2022-08-05
Replace react-codemirror6 with @uiw/react-codemirror

felixroos:
Because of https://github.com/tidalcycles/strudel/issues/143 and  https://github.com/tidalcycles/strudel/issues/125 , I switched the codemirror react package.


felixroos:
fixes https://github.com/tidalcycles/strudel/issues/143

Issue #145, opened 2022-06-25
consider using https://vitest.dev/


felixroos:
it's like jest, but it supports the vite build chain, which we are already using. so it supports esm out of the box + snapshot testing.

felixroos:
Fixed with #175

PR #175, opened 2022-08-06
replace mocha with vitest

felixroos:
- refactored all the tests to vitest syntax
- tune test not yet refactored
- blocked by https://github.com/vitest-dev/vitest/issues/1807


felixroos:
fixes https://github.com/tidalcycles/strudel/issues/145

felixroos:
waiting for vitest@0.21.1 https://www.npmjs.com/package/vitest

felixroos:
i have now refactored all tests to vitest + now using the snapshot feature instead of the custom solution. unfortunately, it's actually slower than mocha by ~50% so I am a bit disappointed.. it seems vitest has had a performance drop lately, so hopefully this will get better soon

felixroos:
i think it's still better, because the dev mode seems much quicker + the snapshot feature alone will be of great use. also, if we decide to move to typescript later, this will automatically work. I'll merge this

Issue #184, opened 2022-08-14
Providing too much resolution to early or late results in dropped notes


bpow:
This may be a case of "well don't do that then", but hopefully can be addressed.

I've noticed that if there are a lot of significant digits to the argument of `late` or `early`, then some notes end up being dropped. A quick example that shows this is:

```js
sine.range(0, 12).late(Math.PI).segment(12).scale('C major').pianoroll()
```

Running this in the REPL results in some notes not being played as well as not being displayed on the piano roll (there should be 12 notes per cycle, but some end up being left out).

So far, I think somehow the Number to Fraction to Number conversion may be messing things up, resulting in whole / part components of haps with some fractions with very large numerators/denominators.

Running:
```js
sine.range(0, 12).early(Math.PI).segment(12).queryArc(0, 1).map(e => e.show())
```

Results in:
```js
[
    "(0/1 -> 1/12, 0/1 -> 355271367880050/4263256414560601, 11.480137265535536)",
    "(1/12 -> 1/6, 1/12 -> 1/6, 11.967422385140505)",
    "(1/6 -> 1/4, 947390314346802/5684341886080801 -> 888178419700124/3552713678800501, 10.855741495751651)",
    "(1/4 -> 1/3, 1/4 -> 1/3, 8.442968593921858)",
    "(1/3 -> 5/12, 1421085471520202/4263256414560601 -> 5/12, 5.3756042302161084)",
    "(5/12 -> 1/2, 2368475785867002/5684341886080801 -> 1776356839400250/3552713678800501, 2.475546208781361)",
    "(1/2 -> 7/12, 1/2 -> 7/12, 0.5198627344644546)",
    "(7/12 -> 2/3, 2486899575160352/4263256414560601 -> 2/3, 0.032577614859495174)",
    "(2/3 -> 3/4, 3789561257387202/5684341886080801 -> 2664535259100374/3552713678800501, 1.1442585042483464)",
    "(3/4 -> 5/6, 3/4 -> 5/6, 3.5570314060781407)",
    "(5/6 -> 11/12, 3552713678800502/4263256414560601 -> 11/12, 6.62439576978387)",
    "(11/12 -> 1/1, 5210646728907402/5684341886080801 -> 3552713678800500/3552713678800501, 9.524453791218637)"
]
```

Running with early/late shifts with fewer significant digits (e.g., does not result in this issue of dropped notes (at least not after listening for a while...

felixroos:
[...]

PR #185, opened 2022-08-14
change "stride"/"offset" of successive degradeBy/chooseIn

bpow:
This avoids (without fixing) #184).

Note that the offset in tidal is currently 0.0001, but using this
would result in excess correlation between sequential uses of
random, which would manifest in failure of the 'supports the randome
choice operator with nesting' test. So I just used 0.0002 instead


bpow:
This had come up because of some weirdness that @felixroos noticed after accepting #165.

It seems like maybe I was trying to be too clever, in avoiding the correlation of nearby samples inherent in the current rand implementation, by offsetting successive random strides into the stream by an irrational number (well, Math.PI isn't really irrational, but it has a lot of significant digits). This triggers some weirdness which I've reported in issue #184 .

The quickest fix is to use something else for the stride offset of a new random stream for each choose/degradeBy. If we back off to what it was before and in tidal (0.0001), then there's enough correlation between successive random stream (e.g., a degradeBy and subsequent chooseIn), that there is some very evident non-randomness showing up in some very simplistic statistical tests. More importantly, this correlation is noticeable to users.

So this may be a temporary fix (if/until the `rand` implementation is altered), but changing the offset to be 0.0002 both prevents dropped notes and avoids the most obvious correlative behavior (at the downside that the 500th random stream will end up being an exact offset of the 1st random stream by an exact cycle, instead of the 1000th).

felixroos:
could this be fixed by switching the call order to segment -> early ? see https://github.com/tidalcycles/strudel/issues/184#issuecomment-1214332749

felixroos:
thanks for the fix! I think this is fine for now when https://github.com/tidalcycles/strudel/issues/184 is done, it might be time to look at this again

PR #193, opened 2022-08-20
add support for legato to the mininotation

bpow:
I just found out about `legato`!

It's pretty cool, and it could be a lot easier to express staccato that way as opposed to shortening durations and following up with rests of exactly the right size. However, there's no support for it in the mininotation, so here's a proposal.

For instance, you could do something like `"[a4 c5 e5]=0.5 [a5 e5 c5]"`

**Note:** the mini-notation is rapidly running out of sigils to use for different operators, I went with `=` for now, but that might bear some discussion since `=` has been mentioned as a possibility to use for notes held over after a tie. It looks like maybe `&` and `$` may be still available...

### DIFF
--- a/packages/mini/krill-parser.js
+++ b/packages/mini/krill-parser.js
@@ -195,20 +195,22 @@ function peg$parse(input, options) {
   var peg$c21 = "*";
   var peg$c22 = "%";
   var peg$c23 = "?";
-  var peg$c24 = "struct";
-  var peg$c25 = "target";
-  var peg$c26 = "euclid";
-  var peg$c27 = "slow";
-  var peg$c28 = "rotL";
-  var peg$c29 = "rotR";
-  var peg$c30 = "fast";
-  var peg$c31 = "scale";
-  var peg$c32 = "//";
-  var peg$c33 = "cat";
-  var peg$c34 = "$";
-  var peg$c35 = "setcps";
-  var peg$c36 = "setbpm";
-  var peg$c37 = "hush";
+  var peg$c24 = "=";
+  var peg$c25 = "struct";
+  var peg$c26 = "target";
+  var peg$c27 = "euclid";
+  var peg$c28 = "slow";
+  var peg$c29 = "rotL";
+  var peg$c30 = "rotR";
+  var peg$c31 = "fast";
+  var peg$c32 = "scale";
+  var peg$c33 = "legato";
+  var peg$c34 = "//";
+  var peg$c35 = "cat";
+  var peg$c36 = "$";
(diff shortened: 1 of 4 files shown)


yaxu:
Thanks for this, however legato is moving from 'context' metadata to data https://github.com/tidalcycles/strudel/issues/111

There are many other aspects of a sound that we could want to pattern in this way, and as you say we're running out of operators.

`note "c e f g" # legato "0.5 0.25 0.2 0.2"` is fiddly if you want to adjust every note but the same can be said for any other kind of effect.

A more general but yet-to-be-implemented approach similar to what is seen in tidal would be to use `:` as a delimiter for making a pattern of lists. Then doing something like:

```js
"[a4 c5 e5]:0.5 [a5 e5 c5]".assign(note, legato)
```

We can build this into some 'compound' controls so e.g. `adsr("0.25:0.5:0.25:0.1")` and `sound("bd:4 sd:2")` works

I think duck typing makes this approach more practical in javascript than it is in haskell.

felixroos:
assigning control params (like legato) to bare values should work with https://github.com/tidalcycles/strudel/issues/288

```js
"[a4 c5 e5]=0.5 [a5 e5 c5]" // { value:'a4', legato:0.5} etc...
.note() // { note:'a4', legato:0.5 } etc...
```

felixroos:
> A more general but yet-to-be-implemented approach similar to what is seen in tidal would be to use `:` as a delimiter for making a pattern of lists. Then doing something like:
> 
> ```js
> "[a4 c5 e5]:0.5 [a5 e5 c5]".assign(note, legato)
> ```
> 
> We can build this into some 'compound' controls so e.g. `adsr("0.25:0.5:0.25:0.1")` and `sound("bd:4 sd:2")` works
> 
> I think duck typing makes this approach more practical in javascript than it is in haskell.


update from the future: the ":" operator exists. The above mentioned `assign` function is prototyped here: https://github.com/tidalcycles/strudel/issues/595 called "as". It can be used to solve the specific problem of this PR using the following syntax:

```js
"[a4 c5 e5]:0.5 [a5 e5 c5]".as("note:clip")
```

note: [clip](https://strudel.cc/learn/time-modifiers#clip) is basically the same as legato (pun intended :P)

I generally like the idea of using ":" instead of adding a new operator, because there is only a limited number of available operators + any operator that would set a key on the hap.value is problematic to use on a hap of plain values. 
Also, the `as` function is very self-explanatory, so you don't need to learn special syntax

yaxu:
`as` now exists, so I think we can close this for now.

Issue #195, opened 2022-09-05
webaudio effects - feature parity with webdirt


felixroos:
It would be nice if webaudio supported all effects in webdirt, missing:

- [x] vowel https://github.com/tidalcycles/strudel/pull/201
- [x] delay, delaytime, delayfeedback https://github.com/tidalcycles/strudel/pull/213
- [x] nudge, unit, cut, loop
- [x] speed with negative  https://github.com/tidalcycles/strudel/pull/222
- [x] coarse, crush, shape https://github.com/tidalcycles/strudel/pull/205
- [ ] accelerate - wont fix for now..

If those are implemented, oscillators and samples can be processed by the same output, using the same effects logic.
Having two different implementations for the same thing requires synchronizing both when things change.
I reimplemented it in the first place because webdirt does not support oscillators + does not expose web audio nodes for future processing.. also because it was a good opportunity to learn some webaudio things

felixroos:
vowel: https://github.com/tidalcycles/strudel/pull/201

felixroos:
turns out webdirt supports these effects too:

- [shape](https://github.com/dktr0/WebDirt/blob/8d57a932970f554b462f8115cc89a2985654dc7e/js-src/AudioWorklets.js#L66)
- [crush](https://github.com/dktr0/WebDirt/blob/8d57a932970f554b462f8115cc89a2985654dc7e/js-src/AudioWorklets.js#L32)
- [coarse](https://github.com/dktr0/WebDirt/blob/8d57a932970f554b462f8115cc89a2985654dc7e/js-src/AudioWorklets.js#L1)

felixroos:
squiz [doc](https://tidalcycles.org/docs/patternlib/tutorials/audio_effects/#squiz) [scd](https://github.com/musikinformatik/SuperDirt/blob/44ef3ef35d6d0cd94405596ca41241c1331ebf09/library/default-effects-extra.scd#L26) [cpp](https://github.com/supercollider/sc3-plugins/blob/main/source/MCLDUGens/MCLDDistortionUGens.cpp)

PR #205, opened 2022-09-15
Coarse crush shape

felixroos:


felixroos:
ref https://github.com/tidalcycles/strudel/issues/195

Issue #58, opened 2022-04-13
Missing tutorial documentation


yaxu:
There's probably quite a few things missing at this point, lets make a list:
* [ ] `chop`
* [ ] `jux`
* [ ] ...

felixroos:
There is now a auto generated version of this list here: https://github.com/tidalcycles/strudel/blob/main/undocumented.json which can be regenerated by via `npm run report-undocumented` , which runs https://github.com/tidalcycles/strudel/blob/main/undocumented.mjs which parses all source files for exports and cross checks them with the jsdoc comments

PR #212, opened 2022-09-17
Even more docs

felixroos:
- overall deduplication + restructuring
- now most of the actual function documentation is in source, the tutorial only references them
- documented some more functions
- marked tone + webdirt as deprecated


felixroos:
ref https://github.com/tidalcycles/strudel/issues/58

PR #213, opened 2022-09-18
Feedback Delay

felixroos:


felixroos:
TODO: delay node lifespan seems limited, either make it global or find out how to know the lifespan in advance

felixroos:
delay now works as a global effect with orbits. they are lazily created once for each orbit. the same approach can be used for other global effects

PR #220, opened 2022-09-23
Fix squeeze join

yaxu:


yaxu:
This seems to fix it, but I'm not sure exactly what's going on.

The inner pattern is sped up and shifted so that cycle one happens within the outer event 'whole'. But by my thinking that would mean that you'd always get events from that first cycle.

I.e. this gives 0.25, 0.75 from the first cycle, rather than 1.25 and 1.75 from the queried cycle.
```js
let p = time.segment(2)._focus(Fraction(1),Fraction(2))

console.log(p.queryArc(1,2).map(x => x.value))
```
However, this works as we want, i.e. we get 1.5 and not 0.75.
```js
let p = time.segment(1).ply(2)
console.log(p.queryArc(1,2).map(x => x.value))
```
So as far as I can tell, it's fixed, but I don't know how - I was fully expecting to have to shift time again before the query of the inner pattern..

yaxu:
The same approach works in Tidal 2.0, so I guess I'll just merge it!

PR #232, opened 2022-10-27
Out by default

felixroos:
This PR marks a little milestone, removing the long standing tone default synth and instead uses the webaudio output `.out()` by default.
- removed all usages of `.out()` as they are now redundant and probably confusing. 
- removed tone and webdirt APIs because webdirt has now been fully replaced and the Tone output never worked in a way that made sense for strudel.
- added a little sugar to automatically recognize primitive hap values of note shape as notes, which keeps all the tutorial snippets as simple as they are + resolves some confusion.


felixroos:
ok to merge?

felixroos:
@yaxu ok to merge?

yaxu:
Looks good except I'm not in favour of defaulting to notes. If I'm understanding it right, `/^[a-gA-G][#b]*[0-9]$/` could match things that aren't notes, like shorthands for fractions (`e` = eighth), vowels, samples indexed by letter etc.

felixroos:
it will only match notes with octave, so `e` wouldn't be matched. Also remember that this only applies to haps with nothing else defined, meaning something like a vowel would not have an effect (no sound is played). the only thing that really could clash are sample names that look like notes, which could even be checked for. e.g. if it is something that is a note but also a sample name, show a warning or an error .. As this is generally rare, and also avoidable when wrapped, I don't think it must be added in this PR, also we might want to wait for https://github.com/tidalcycles/strudel/issues/233

yaxu:
I think it would be good if there was a default octave for notes so you wouldn't need the number, this is handy in tidal.

I don't think notes should be more special than vowels, or any other dimension of sound, especially in the tutorial.

felixroos:
okay I guess I cannot convince you :) then I'll update all the tutorial snippets to always use `note(..)` + throw an error if the output receives a non object hap

yaxu:
Thanks for taking that on board, lets see how it feels

felixroos:
The whole tutorial is now refactored to note calls

PR #239, opened 2022-11-02
Load samples from url

felixroos:
can now use the `samples` function with a url that points to a json file, e.g.

```js
samples('https://gist.githubusercontent.com/felixroos/b9d461966ae1aaa660beea7c61482d21/raw/dirt-samples.json')
```

this allows loading custom sample sets more easily


yaxu:
Wonderful!

PR #243, opened 2022-11-03
snapshot tests on shared snippets

felixroos:
This test runs a snapshot test on all snippets in the database. After adjusting the test runtime (added more mocks) + fixing some old snippets, all 289 snippets now run without error. 

So far, I excluded the test from the normal test suite, as I want to fire requests from the test suite.

Either this test could be ran manually from time to test a potential breaking change on a larger set of snippets. Or we run it by default on a database dump that can be updated from time to time.

At a certain point it will be too much work to maintain user snippets (and keep the test green) as more people start using strudel, but for now it seems like a good way to get more data to test on. 

Function specific unit tests are of course still important, but it seems that running tests on real world examples brings out situations that a unit test does not come encounter. 

Another related question is how to handle / communicate the privacy aspect of the snippet sharing function, as user data is stored in the supabase, and with this test that data is also processed. It could also be a legal issue to collect user data without having a gdpr compliant text on the website but I don't know much about these laws


felixroos:
I'll merge this now, added a dbdump. it's not part of the main test suite because it takes quite long, but it should be good to have that snapshot data to test from time to time

PR #265, opened 2022-11-17
fake curried controls, maybe fixes #264

### LINKED ISSUE #264: Curries controls ?
Currently this sort of thing doesn't work:
```js
s('bd').every(3,n(2))
```
This feels surprising, `s('bd').x(2)`, `fast(2, x('bd'))` and `s('bd').every(3, x(2))` work when x=`fast`, but not when it's a control like `n`.

I don't think we can fix this with currying. Maybe it's up to `every` and friends to check whether its last argument is a function or a pattern, and in the latter case combine it with `setIn`.. 

That would imply support for e.g. `s('bd').every(3,setOut.n(2))` as well.

yaxu:


felixroos:
not sure if we have talked about this one in chat, but is this still relevant, or an alternative solution to https://github.com/tidalcycles/strudel/pull/390 ?

yaxu:
Yes a previous attempt, but might eventually be useful if I come full circle..

PR #274, opened 2022-11-30
Simplify how/what matrix into separate methods, build exports from pattern.mjs dynamically

yaxu:


packages/core/controls.mjs
```diff
@@ -4,7 +4,8 @@ Copyright (C) 2022 Strudel contributors - see <https://github.com/tidalcycles/st
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more details. You should have received a copy of the GNU Affero General Public License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { Pattern, sequence } from './pattern.mjs';
```
felixroos, line 7: wouldn't it be simpler if we kept those export as they are and only export an additional `pattern` object? this way we can still have simpler imports (left side), while having a monster object to use as a scope.

packages/core/draw.mjs
```diff
@@ -4,7 +4,11 @@ Copyright (C) 2022 Strudel contributors - see <https://github.com/tidalcycles/st
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more details. You should have received a copy of the GNU Affero General Public License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { Pattern, getTime, State, TimeSpan } from './index.mjs';
```
felixroos, line 7: same here

packages/core/index.mjs
```diff
@@ -4,27 +4,36 @@ Copyright (C) 2022 Strudel contributors - see <https://github.com/tidalcycles/st
...
 */
 
-import controls from './controls.mjs';
-export * from './euclid.mjs';
 import Fraction from './fraction.mjs';
+
+import drawLine from './drawLine.mjs';
+import gist from './gist.js';
 import { logger } from './logger.mjs';
-export { Fraction, controls };
```
felixroos, line 11: we could keep all of these

packages/core/index.mjs
```diff
@@ -4,27 +4,36 @@ Copyright (C) 2022 Strudel contributors - see <https://github.com/tidalcycles/st
...
-export * from './pianoroll.mjs';
-export * from './ui.mjs';
-export { default as drawLine } from './drawLine.mjs';
-export { default as gist } from './gist.js';
+// Fraction: Fraction, drawLine: drawline, gist: gist, logger: logger,
+
+import controls from './controls.mjs';
+import pattern from './pattern.mjs';
+
+import * as draw from './draw.mjs';
```
felixroos, line 17: and still add these

felixroos (changes requested):
generally speaking, I think changing all import structures in all packages is a bit heavy.. we could keep everything as is and just add the additional monster object for the scope.. in some situations it might even save memory to do so, because you might not want the monster object, only some imports, which would then be tree shaked....

packages/core/controls.mjs
yaxu: Hm, but then there's no clear reasoning for why a function is exported directly, or only available in the pattern object, and it might appear and disappear based on how its implemented internally.

felixroos: hm yes all the dynamically generated functions would only appear in the big object, the rest could be exported separately (in addition to being in the big object). this just simplifies importing 90% of all things, because you don't have to always do `const { x, y, z } = core`. also there won't be breaking changes everywhere. We could even go so far to write a script that exports all the matrix functions in the root scope..

yaxu: Hm I think I'd rather just scrap all this and generate sourcecode then. We just do everything with methods in pattern.mjs, list or mark the functions we want to promote to curried toplevel functions, and generate that as a separate source file called something like `pattern-static.mjs`. We then add a build script like we do in the mininotation package.

yaxu: There'd then be a question about whether to generate controls.mjs to be exported directly as well. It's quite nice to have them organised a bit separately, but if they all end up in the same scope anyway.. To be consistent, best to either put everything in one object, or have everything directly exported?

felixroos: it would still be practical to have a non generated big object with all things that should go into the eval scope. for all internal imports this should be fine. i see the source code generation more as a way to provide a simpler import api for library users

yaxu: I think almost all the non-methods should be either dynamically or statically generated. Otherwise we get this sort of thing which difficult to maintain:
https://github.com/tidalcycles/strudel/blob/485a96962ad4a8057a8387f14ae1fa42755e9873/packages/core/pattern.mjs#L1822-L1852

One possible advantage of statically generated rather than dynamically generated code is that we can generate documentation along with it, so that conventional documentation tools work.

felixroos: > I think almost all the non-methods should be either dynamically or statically generated. Otherwise we get this sort of thing which difficult to maintain:

agreed

> One possible advantage of statically generated rather than dynamically generated code is that we can generate documentation along with it, so that conventional documentation tools work.

jsdoc also works in thin air, when used with @name and @memberof. So this should be no limiting factor

felixroos: but i'd say static written code is much easier to read

yaxu: If the main argument is that a monster object is a breaking change, I think that's fine. Having a single, consistent way to export stuff means less breaking in the future.. and this PR already fixes all the internal breaks I could find.

For the memory usage argument, I don't know how javascript deals with memory but guess we could try it..

yaxu: > but i'd say static written code is much easier to read

The problem is there are inconsistencies and a lot of missing toplevel functions. I don't really want to manually fix this as it'd be an ongoing headache to keep it up-to-date.

Unless we switched things around and wrote everything as a toplevel module function, with a decorator (via babel?) to add it to the class as a method.

felixroos: talking about memory i meant tree shaking, which means only the things that are imported are used. Breaking changes is one argument, but the main argument is that default exporting big objects is generally considered to be an "antipattern".. it's more difficult to auto import stuff + you always have to write an extra statement by hand.. i heavily use code completion which just automatically imports the stuff i am using, which does not work this way.. some written words about this: https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad

felixroos: i still use default exports btw but only when a file exports a like function / value etc..

yaxu: Here's the Tidal equivalent of controls.mjs: https://github.com/tidalcycles/Tidal/blob/main/src/Sound/Tidal/Control.hs

It's autogenerated and reads the same as when it was handmade.. just more consistent and with fewer bugs https://github.com/tidalcycles/Tidal/blob/main/src/Sound/Tidal/Control.hs

felixroos: > The problem is there are inconsistencies and a lot of missing toplevel functions. I don't really want to manually fix this as it'd be an ongoing headache to keep it up-to-date.

yes, I also want this to be autogenerated (I mean autogenerated on an object, not autogenerated with a script that outputs code)

> Unless we switched things around and wrote everything as a toplevel module function, with a decorator (via babel?) to curry it and add it to the class as a method.

This is (without huge hassle) only practical as a build step for distribution. Inside the code base, this should be pretty hard to do, as we have to integrate it into the dev server / hot reloading etc.. imo not worth it just to be able to import things from dynamically generated objects in one line less.

yaxu: Ok I'm happy to accept that the monster object export isn't ideal.

That's a shame about lack of decorator but I guess that's only syntactic sugar anyway

```js
export class Pattern {
  // .. bare minimum goes here
}

export const every = _register(function(n, func, pat) {
    // ... definition of every
  }
);
```

Where `_register` adds the method to the class, and returns a curried version.. hopefully sorting out the composition stuff as well. Then we get both the exports and the class of methods.

We have to define each of the operators by hand but that's fine now that it's of the form add.mix rather than addMix.

felixroos: that looks like it could solve this mess..

felixroos: we need to pass the function name i think (to set `Pattern.prototype[name]` programmatically) 

```js
_register('every', function(n, func, pat) {
    // ... definition of every
  }
);
```

felixroos: I'll have to sleep over this...

yaxu: Ah yes.. 

With this approach we could also add the ability to 'bless' anything into being a pattern, not just the Pattern class.

Will have a think as well. I can extract the matrix howWhat -> how.what change into a separate PR at some point..

yaxu: Ah yes.. 

With this approach we could also add the ability to 'bless' anything into being a pattern, not just the Pattern class.

Will have a think as well. I can extract the matrix howWhat -> how.what change into a separate PR at some point..

yaxu:
Closing in favour of https://github.com/tidalcycles/strudel/pull/285 , which has the monster object removed per above discussion..

PR #283, opened 2022-12-04
Add Csound output with MIDI semantics

gogins:
This PR adds a new Csound output that uses MIDI semantics (MIDI key number instead of frequency, MIDI velocity number instead of normalized amplitude).

I have updated my own build of Csound for WebAssembly to use camel case for all Csound API methods. Hence, there should now be a substantial consistency and interoperability between [my csound-wasm](https://github.com/gogins/csound-wasm) and [the canonical build](https://github.com/csound/csound/tree/master/wasm).

### DIFF
--- a/packages/mini/mini.mjs
+++ b/packages/mini/mini.mjs
@@ -103,11 +103,11 @@ function resolveReplications(ast) {
   });
 }
 
-export function patternifyAST(ast) {
+export function patternifyAST(ast, code) {
   switch (ast.type_) {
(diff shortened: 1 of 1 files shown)


gogins:
There was a typo in csound.mjs line 163 where the word warning needed to be in single quotes.

It was working for me but obviously that is not enough. In future I will run `npm test` before issuing any PRs. I fixed the error, rebased, rebuilt, and npm test passed csound.mjs.

PR #285, opened 2022-12-07
Switch 'operators' from .whatHow to .what.how

yaxu:
Makes sure there are toplevel curried functions for all of them.
Renames util.mod to util._mod, to make room for toplevel 'operator' of that name.


felixroos (approved):
looks good! maybe we can rename _mod to modulo at some point

yaxu:
> looks good! maybe we can rename _mod to modulo at some point

I did! But then changed it to this, as there's a convention for non patterny to have the underscore.. but no strong opinion

PR #286, opened 2022-12-07
Reorganise pattern.mjs with a 'toplevel first' regime

yaxu:


felixroos:
looks promising!

yaxu:
`pattern.mjs` is looking tidier now. Other modules are still using `define` rather than the new `register`, and should probably be updated.

yaxu:
I had a go at tidying up degrade/undegrade in signal.mjs along the same lines but hit the complication that register makes heavy use of `reify`, which applies `pure` which is for discrete values.. this messes things up.
So I propose merging for now and doing further tidying later.

felixroos:
I found a way to get rid of the arity switch case in register! I'd be fine with a merge. We can refactor `makeComposable` / `patternifyX` / `Pattern.prototype.define` in a separate PR

felixroos (approved):

yaxu:
Heh, let me know your t-shirt size and address !

felixroos:
> I had a go at tidying up degrade/undegrade in signal.mjs along the same lines but hit the complication that register makes heavy use of reify, which applies pure which is for discrete values.. this messes things up.
So I propose merging for now and doing further tidying later.

just tried using `register` with degrade(By) / undegrade(By) and it seems to work even with reify. see https://github.com/tidalcycles/strudel/commit/09b15a07c63bd17e388f90b0d7c0287c40a82207

should I just refactor the rest of `signal.mjs` similar to this or is there something messed up I am missing? I also tried to add a noReify option, but skipping reify actually breaks stuff

yaxu:
Yeah I tried the same and think the use of pure was putting a sample and hold on the random signal so it was either all or nothing on a per cycle basis.

yaxu:
We need to reify but with steady rather than pure.

felixroos:
I wonder why it worked with the old patternify function

yaxu:
I'm afk but it's using something custom with applicative.. hmm

PR #291, opened 2022-12-09
Fancy hap show, include part in snapshots

yaxu:
Parts are visualised like:
where whole = part: `0/1 → 1/1`
where they don't: `2/3 ⇜ (3/4 → 1/1) ⇝ 4/3`
where the onset is still intact but offset isn't: `(2/3 → 3/4) ⇝ 4/3`
vice-versa: `2/3 ⇜ (5/4 → 4/3) `

The whole thing looks like:
```
[ 2/3 ⇜ (1/1 → 5/4) ⇝ 4/3 | note:64.01075614638685 cutoff:688.232561769837 resonance:12 gain:0.5 decay:0.16 sustain:0.5 delay:0.2 room:0.5 pan:0.6 s:piano clip:1 ]
```


felixroos (approved):
i like the new whole / part format!

PR #303, opened 2022-12-13
Updated csoundm to use the register facility .

gogins:
I'm trying this pull request again. If you don't want this just say so.

The logic you had for using an external instance of Csound was not suited to my use cases, and would have failed in use. The external Csound will, usually, already have compiled an external orchestra and be running, so the csound.mjs module should not call compile or start on an external instance of Csound.

I updated the csoundm function to use the new 'register' function for Patterns. I also updated csoundm to convey the Strudel signals in the same way as the existing csound function.

### DIFF
--- a/packages/csound/csound.mjs
+++ b/packages/csound/csound.mjs
@@ -79,21 +79,25 @@ function eventLogger(type, args) {
 
 async function load() {
   if (window.__csound__) {
-    // allows using some other csound instance
+    // Allows using some other csound instance.
+    // In that case, the external Csound is responsible
+    // for compiling an orchestra and starting to perform.
+    logger('[load] Using external Csound', 'warning');
     _csound = window.__csound__;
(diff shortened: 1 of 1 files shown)


felixroos:
I'll merge this now, though it might make more sense to find a less specific abstraction in the future, making this one obsolete, as outlined in https://github.com/tidalcycles/strudel/pull/283#issuecomment-1345631658 . also, a solution for key value mappings is still missing, as described here https://github.com/tidalcycles/strudel/discussions/270#discussioncomment-4287232 , an alternative implementation might use channels with unique names.

gogins:
I am re-examining your comments, and trying to envision what your use cases
are. More later.

Best,
Mike

-----------------------------------------------------
Michael Gogins
Irreducible Productions
http://michaelgogins.tumblr.com
Michael dot Gogins at gmail dot com

PR #315, opened 2022-12-20
Astro build

felixroos:
This new build setup aims to combine the repl and the tutorial, using astro to build. it also supports file based routing. so far, I started with https://github.com/withastro/astro/tree/latest/examples/docs + began refactoring the styles to tailwind.


felixroos:
this has now reached a state where it could already to replace the existing build. The docs are still not perfect on mobile though + algolia search is still missing

PR #341, opened 2023-01-02
Patternify euclid, fast, slow and polymeter step parameters in mininotation

### LINKED ISSUE #72: Support patterning euclidean values in the mininotation
e,g `"a3(<3 5>,8,<0 2 4>)"`

yaxu:
e.g. 

```
"x(<3 5>,<8 8 16>)"
"{a b, c d e}%<4 5>"
"[a b]*[3 8 16]"
```
Fixes #72


yaxu:
I haven't patterned the parameters for `!`, `@` or `?`. The first two would be difficult (maybe impossible) to implement. The second would be possible but there is a hitch because the parameter is optional. We could say `4? <0.5 0.2>` are separate events and `4?<0.5 0.2>` is one due to whitespace. But we can't do that without changing how `slice` consumes whitespace.

PR #350, opened 2023-01-05
Support for multiple mininotation operators

### LINKED ISSUE #176: Only one "modifier" is allowed per "slice" in mininotation
This could be reasonably tagged as "compatibility" because the behavior differs from the mininotation in tidal. That also means it is related to #30. I am not sure how much the lack of complete compatibility with tidal should be considered a strudel issue here, though, since I don't think the tidal behavior is very well defined.

Tidal's mininotation allows (some but not all) combinations and orderings of modifiers to a single "slice". I'm using the terminology from within the peggy grammar, where a "slice" would be something like `a5` and the modifier something like `@2` or `?` or `(5,8)`.

For instance, tidal allows/parses `"[c(5,8),e*2(5,8)]"`, which results in a pattern like:
```
|c-..c-c-..c-c-..|c-..c-c-..c-c-..|c-..c-c-..c-c-..|c-..c-c-..c-c-..
|ee..eeee..eeee..|ee..eeee..eeee..|ee..eeee..eeee..|ee..eeee..eeee..
```

In the case of the repeat and euclidean operators, they cannot be put in the other order (e.g., `e(5,8)*2` parses neither in tidal nor strudel).

Tidal does allow some combinations of modifiers to have different combinations (`c?*2 d` and `c*2? d` both parse in tidal and have different outputs). They essentially end up getting interpreted the same as `[c?]*2` and `[c*2]?` respectively (that is, the first modifier binds "more tightly").

For what it is worth, both tidal and strudel can handle the more explicit formulations of for combining euclidian and repeat modifiers with extra square brackets, like [`"[[e4*2](5,8),[c3(5,8)]*2]"`](https://strudel.tidalcycles.org?YD2MRLffOCRV).

yaxu:
Fixes #176


yaxu:
Stuff like `bd(3,8)*2` seems to be working, although events are coming out in a different order for some strange reason, breaking a couple of tests.
Also could do with some new tests

PR #365, opened 2023-01-09
Update mini-notation.mdx

yaxu:
`x!n` is no longer an alias for `[x*n]@n` and the behaviour is a bit different. Probably best to remove this line in case it confuses someone?


yaxu:
(Not sure though, maybe it's still a useful analogy.)

felixroos:
maybe good to remove it.. it should be clear what it does

PR #367, opened 2023-01-11
Fixed formatting and performed npm run check.

gogins:


felixroos:
sorry for the late response, this seems to already have been [fixed on the upstream](https://github.com/tidalcycles/strudel/blob/1015bb0fdcb044cf3a3333156a38341d393a19e3/packages/webaudio/sampler.mjs#L128). thanks anyway 🥇

PR #368, opened 2023-01-11
Support functional composition WIP

yaxu:
With this PR, patterns like
```js
s("bd sd").every(3, x => x.speed("2").fast(2))
```
can be written like
```js
s("bd sd").every(3, speed("2").fast(2))
```


yaxu:
Getting non-controls working is a bit harder:
```js
s("bd sd").every(3, fast(2).iter(4))
```
because `fast(2)` isn't a pattern. I've added a `__compose` method to it, but that isn't useful because it doesn't have all the pattern methods to compose with it.

We could do that by adding loads of stub functions that do the composing, one for each pattern method.

Alternatively we could somehow make functions that are patterns, or make 'dummy' patterns.

packages/core/util.mjs
```diff
@@ -149,6 +151,25 @@ export function curry(func, overload, arity = func.length) {
       const partial = function (...args2) {
         return curried.apply(this, args.concat(args2));
       };
+
+      if (args.length == arity - 1) {
+        // The penultimate arg.. so add some composition magic
+        // TODO - To make this useful, we also need to add stub functions for every pattern method to
+        // do the actual composing
+        for (const r of Pattern.__registered) {
```
felixroos, line 159: is __registered already fully populated at the time of executing this? this reminds me of the problem I *tried* to solve with the now extinct Pattern.prototype.bootstrap

packages/core/pattern.mjs
```diff
@@ -1368,9 +1372,32 @@ export function register(name, func) {
       return func(...args, pat);
     };
     mapFn = curry(mapFn, null, arity - 1);
-    return right.reduce((acc, p) => acc.appLeft(p), left.fmap(mapFn)).innerJoin();
+
+    // Don't use applicative for arguments that a) have a '__compose' function and b) are
+    // marked as being a higher order function parameter
+    function app(acc, p, i) {
+      if (f_params != null && f_params.length > i + 1 && f_params[i + 1] && '__compose' in p) {
```
felixroos, line 1379: maybe things would be easier to define if f_params would be an array of indices? like `register('xx', .., [1]) ` tells that the argument with index 1 is higher order.

packages/core/controls.mjs
```diff
@@ -819,13 +819,23 @@ const _setter = (func, name) =>
   };
 
 generic_params.forEach(([type, name, description]) => {
-  controls[name] = (...pats) => _name(name, ...pats);
+  controls[name] = function (...pats) {
+    const result = _name(name, ...pats);
+
+    // Add a function for composing this control with another pattern
+    result.__compose = function (pat) {
```
felixroos, line 826: shouldn't this be done inside _name, so createParam will get the same treatment?

packages/core/util.mjs
yaxu: Not sure, I was hoping this wouldn't be a closure so would always be up to date.

But I'm planning on just returning a dummy pattern with `__compose` as the function instead of doing this.

packages/core/pattern.mjs
yaxu: I wondered about this too but think there will be other things we want to say about the parameters, like the default alignments ref #380.

yaxu: > I wondered about this too but think there will be other things we want to say about the parameters, like the default alignments ref #380.

I mean in that case a list of objects would probably be best

felixroos: jep.. makes sense

yaxu:
A thought..
`n(3)`, `pure(3)` and `sine` are different kinds of patterns.

If we have `s("bd").every(3, n(3))` then maybe we should also support `pure(3).every(3,sine)`, and especially `s("bd").every(3, pure(3).n()).

If that makes sense, then it means that _every_ pattern should also be able to be treated as a composable function.

We can add an `as_function` method or getter that defaults to `x => x.set(this)`.

The problem then is what to do for higher order methods like `.every(n, func)`. The second parameter accepts a function, but is patternified. If every pattern is also a function, then we don't know whether to apply it as a function, or use applicative to apply it as a pattern-of-functions.

In other words, we want to support this:
```js
s("bd sd").every(3, sequence(rev, fast(2)))
```
but also this:
```js
s("bd sd").every(3, n(2)))
```
We want to apply the first as a pattern of functions, and the second as a function of patterns. Currently we do the former if the pattern doesn't have a `_compose` function, but if all patterns have one of those..

I fear the answer is that we have to start making a type system..

yaxu:
So the problem is that 
```js
sequence(rev, fast(2))`
```
could either be treated as `x => x.set.sequence(rev, fast(2))` or `x => x.apply(sequence(rev, fast(2)))`

There's no way of knowing without implementing some kind of type system.

Strudel currently even lets you mix functions with plain values:
```js
sequence(rev, 3)`
```
.. which shows that there's currently really nothing about a pattern that helps us decide whether to treat it as a pattern or as `a x => x.set(pattern)` function.

So maybe in the end we just have only support explicit functions e.g.

```js
x.every(3, set.n(3))
```
and through composition
```js
x.every(3, set.n(3).fast(3))
```
but not
```js
x.every(3, n(3))
```

felixroos:
What follows is me trying to make sense of this..

> If we have s("bd").every(3, n(3)) then maybe we should also support pure(3).every(3,sine), and especially s("bd").every(3, pure(3).n()).

What would `pure(3).every(3,sine)` event mean? So far we have no chainable sine method, also no concept of what it would mean. The same goes for `pure`. I can hardly think about this stuff without seeing any real use case..

So far my mental model has been to just get rid of `x=>x.` and keep the end result the same. 

> So the problem is that `sequence(rev, fast(2))` could either be treated as `x => x.set.sequence(rev, fast(2))` or `x => x.apply(sequence(rev, fast(2)))`.

- The former seems to not make any sense in that context as you cannot set a function!?
- The latter seems to be the thing that makes sense, example: [https://strudel.tidalcycles.org?wbmLefru_f4g](https://strudel.tidalcycles.org/?wbmLefru_f4g)

> So maybe in the end we just have only support explicit functions e.g. `x.every(3, set.n(3))`

So that would be equivalent to `x.every(3, x=>x.n(3))` ?

> and through composition `x.every(3, set.n(3).fast(3))`

And that to `x.every(3, x=>x.n(3).fast(3))` ?

> but not `x.every(3, n(3))` [..] There's no way of knowing without implementing some kind of type system.

I am not sure if this is a problem of a type system, as tidal also uses the "#" for setting / union (is that correct?)

Do I understand you right that if we say `x.every(3, n(3))` is invalid, we still could support everything we want?
So could you say "set" would then just be the replacement for `x=>x` ? Please tell me if I am completely off the tracks..

Another thing that always was in the back off my mind is that we could simplify this whole mess and just use the transpiler to add the x=>x. and call it a day...

felixroos:
on a sidenote: in some cases it is actually pretty useful to have the `x` to do things, like https://strudel.tidalcycles.org?RAh7ZNOs0Il_ though this might be rare

yaxu:
Sorry, I meant `x => x.set(sequence(rev, fast(2)))`.

`pure(3).every(3, sine)` would do `pure(3).every(3, x => x.set.sine)`

Remember that the .set is implied, so we are really trying to get rid of `x => x.set`. My feeling is that we can't do that without introducing ambiguity that can only be resolved by strict types. But by getting rid of `x => x` and keeping the `.set` it's not ambiguous any more.

I can probably be clearer when I'm not typing on my phone while drinking.

yaxu:
Ok a proper reply..

> What would `pure(3).every(3,sine)` even mean? So far we have no chainable sine method, also no concept of what it would mean. The same goes for `pure`. I can hardly think about this stuff without seeing any real use case..

Just as `n(3).every(3, n(5))` would replace the pattern every third repetition, so would `pure(3).every(3, sine)`. Here `n(5)` is shorthand for `x => x.set(n(5))` and `sine` is shorthand for `x => x.set(sine)`.

So here `n(3)` and `sine` can behave either like a pattern or a function, depending on the context. The problem is that we can't always know from the context whether we want the object to behave like a function or a pattern, like in the case of the second parameter to `every`.
 
> > So maybe in the end we just have only support explicit functions e.g. `x.every(3, set.n(3))`
> So that would be equivalent to `x.every(3, x=>x.n(3))` ?

Yep!

> > and through composition `x.every(3, set.n(3).fast(3))`
> And that to `x.every(3, x=>x.n(3).fast(3))` ?

Yep!

I think this is clearer than missing off the default `set.`, and is clearer when switching to `add`, `add.squeeze` etc.

> > but not `x.every(3, n(3))` [..] There's no way of knowing without implementing some kind of type system.
> 
> I am not sure if this is a problem of a type system, as tidal also uses the "#" for setting / union (is that correct?)

Well due to the type system you can't do something ambiguous like `every 3 (n 3)` in Tidal/Haskell, you have to do `(# n 3)` (for `set.n(3)`) or `(n 3 #)` (for `set.out.n(3)`) etc. 

> Do I understand you right that if we say `x.every(3, n(3))` is invalid, we still could support everything we want? So could you say "set" would then just be the replacement for `x=>x` ? Please tell me if I am completely off the tracks..

Yep! `set` is the replacement for `x=>x` and more specifically `x=>x.set`. It's a bit more to type but I think is easier to reason about.

felixroos:
thanks for clarification! So `set` would then completely replace the pattern on the left? Basically like `.every(2, x=>s('bd')` ? example: https://strudel.tidalcycles.org/#cygiYmQiKS5sYXN0T2YoNCwgeD0%2Bcygic2QiKSk%3D in this case, the pattern on the left is completely ignored every second time, whereas `.every(2, x=>x('bd'))` only edits the pattern on the left. I think both functionalities have their place, the latter probably being a good default

PR #379, opened 2023-01-16
Add tidal-drum-patterns to examples

urswilke:
As discussed [here](https://github.com/tidalcycles/strudel/discussions/373#discussioncomment-4691189) this is a first draft to merge the tidal-drum-patterns to the main folder of the examples in the docs.


felixroos:
hey! thanks for looking into this. could you add these to a separate file? tunes.mjs is used for the shuffle function in the repl. having this large number of drum patterns in there will make it really unlikely that the shuffle will pick a tune that is not a drum pattern..

urswilke:
yes sure, should be fixed now. And I felt a bit bad to put all these patterns in your file, which really has some beautiful tunes :)

felixroos:
should I just merge this and then add them to a dedicated page in a separate PR?

urswilke:
that sounds great

PR #392, opened 2023-01-24
Update webaudio worklets to handle webpack support

emptyflash:
Having some issues in flok supporting strudel since it uses webpack:
https://github.com/munshkr/flok/pull/162#issuecomment-1401322986

### DIFF
--- a/packages/webaudio/webaudio.mjs
+++ b/packages/webaudio/webaudio.mjs
@@ -13,7 +13,7 @@ const { Pattern } = strudel;
 import './vowel.mjs';
 
 // Workaround for handling webpack https://github.com/webpack/webpack/issues/11543
-const workletsUrl = new URL('./worklets.mjs?url', import.meta.url);
+const workletsUrl = new URL('./worklets.mjs', import.meta.url);
 
 // export const getAudioContext = () => Tone.getContext().rawContext;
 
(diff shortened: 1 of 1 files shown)


felixroos:
thanks! have you tested if this still works with vite?

emptyflash:
I thought I had, but seems like it isn't working now. I think the url query param trick was causing issues, seems to be working on the website repl with that removed.
However, `vite-vanilla-repl` isn't working with the worklet effects. I tested that on main, and seems to be broken there too.

felixroos:
this is now superseded by https://github.com/tidalcycles/strudel/pull/396 which adds proper builds (including the webaudio package). The builds are now free from bundler specific syntax + also contain a CJS version. The audio init is now also async, which means the start callback can wait for the worklets to be loaded. See [this codesandbox](https://codesandbox.io/s/zen-shannon-l64v0n?file=/src/index.js). I also adapted the vanilla example to a [sandbox](https://codesandbox.io/s/strudel-vanilla-repl-24i0wk?file=/src/index.js). I hope this now works smoothly with webpack. Let me know how it goes!

emptyflash:
@felixroos awesome, I'll try it out, thanks!

munshkr:
@emptyflash Strudel packages upgraded on the `add-strudel` branch, build is working now! Thank you both for the help !

PR #393, opened 2023-01-26
Notes are not essential :)

yaxu:


yaxu:
I think it would be better if sound came first, as the more universal concept?

felixroos:
fine with me, although "some" sounds a bit like an understatement. but I don't have any strong opinion on that

yaxu:
Ok changed to "Pitches are an important building block in many musical traditions"

PR #396, opened 2023-01-30
proper builds + use pnpm workspaces

felixroos:
this PR adds proper builds to each package. So far, the source files were packaged as is (ES Modules / ESM), resulting in problems for CommonJS (CJS) based bundlers, see https://github.com/tidalcycles/strudel/discussions/381 

With this PR, each package has a `build` task, which builds each `index.mjs` as `dist/index.mjs` (ESM) + `dist/index.js` (CJS).

In the process, I refactored the monorepo from [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) to [pnpm workspaces](https://pnpm.io/workspaces). This change was necessary to ensure a smooth package resolution. 

I'll try to explain: To tell a bundler which file is the entry file, the `package.json` has the fields `main` (CJS)  and `module` (ESM), see [explanation](https://esbuild.github.io/api/#main-fields), also [this stackoverflow](https://stackoverflow.com/questions/42708484/what-is-the-module-package-json-field-for). 

The problem: The `main` field is also used by the dev server to resolve packages, see [bare module import](https://vitejs.dev/guide/features.html#npm-dependency-resolving-and-pre-bundling). 

Because of that, using the unbuilt `index.mjs` as `main` allows the dev server to instantly (hot) reload on change (it already does that), without the need to rebuild the whole package. Also, the tests can run without needing to build in advance.  

If the `main` *would* contain the built CJS version, each package would need an additional `watch` task that checks for file changes and reruns the build. This is potentially possible with [workspace watching](https://lerna.js.org/docs/features/workspace-watching), but slows down development + further complicates the setup. Also, the build needs to run before testing to make sure the `main` files are there (adding builds to git is not a nice option).

The rescue comes with [pnpm](https://pnpm.io/) and its ability to overwrite `main` and `module` on publish, using the `publishConfig` field, see [pnpm docs](https://pnpm.io/package_json#publishconfig). Although npm also supports `publishConfig`, it does *not* support setting `main` / `module`, see [npm docs](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#publishconfig). pnpm has additional features that simplify working with monorepos. For example, I was able to find some wrong dependencies, which npm did not tell me about + pnpm is much faster. The only downside is that pnpm requires an additional install via `npm i pnpm -g` + using npm won't work.

I have already tested the publishing, releasing 0.6.x versions: https://www.npmjs.com/search?q=%40strudel.cycles
Finally, using the strudel packages should be possible in all common scenarios, e.g. using [codesandbox](https://codesandbox.io/s/strudel-core-test-forked-9ywhv7?file=/src/index.js).

Also, I've started to update the core examples, the rest will follow.


m-onz:
Nice 1.

Issue #408, opened 2023-02-04
Documentation of envelopes


tedthetrumpet:
There's a link in the reference to '[More info about envelopes](https://strudel.tidalcycles.org/learn/synths-samples-effects/#envelope)' that goes nowhere.

I'd particularly like to request better documentation for using envelopes with _samples_. This is important for making nice crossfading drones or proper granular synthesis.

felixroos:
the envelope params are now on the effects page with https://github.com/tidalcycles/strudel/pull/409
envelopes work the same for synths and samples.
For samples, it make sense to add a separate explainer sorting out `clip`, `cut` , hap durations and envelopes, which can all be used in non obvious ways to control the sample envelope

felixroos:
the envelope doc is now deployed: https://strudel.tidalcycles.org/learn/effects/#amplitude-envelope

PR #409, opened 2023-02-05
improve effects doc

felixroos:
- add envelope docs to effects page
- group effects by "Filters", "Amplitude Envelope", "Dynamics", "Panning", "Waveshaping" and "Global Effects"
- use more logical filter params for doc + make other ones synonyms


felixroos:
ref https://github.com/tidalcycles/strudel/issues/408

PR #413, opened 2023-02-05
google gtfo

felixroos:
remove google fonts, they were not even used.. (gtfo = get the fonts out)


felixroos:
fixes https://github.com/tidalcycles/strudel/issues/412

PR #415, opened 2023-02-05
basic offline cache

felixroos:
added [idb-keyval](https://www.npmjs.com/package/idb-keyval) to store loaded samples to indexDB as base64 data urls.
Once a sample is loaded, it will never be loaded from the network again, unless the indexDB is deleted (which only happens when the user click clear site data, see https://stackoverflow.com/questions/32319691/how-are-unused-indexeddb-databases-cleaned-up).

this means as soon as a particular session was played and loaded, it automatically works offline! i tested this already :)

this could potentially be extended by adding a button "save samples to disk" which runs a task that loads all (!) samples of the default library (or maybe any given strudel.json / sample map). not sure about that though, because that task would fire a ton of requests.

another question is how to handle  if a sample is changing while remaining under the same url.. maybe this doesn't need to be handled because clear site data would do the job..

it could be nice to show which samples are loaded offline in the samples tab though.. potentially adding a little delete button to clear it would not be big deal then


felixroos:
also it might be worth testing what happens with a lot of samples in indexDB.. there might be a limit to hit

felixroos:
superseded by https://github.com/tidalcycles/strudel/pull/419

Issue #182, opened 2022-08-13
add sound registry to integrate soundfonts with `.s`


felixroos:
instead of (or in addition to) `.soundfont`, soundfont presets should be usable via `.s`, integrating with `.out`.
When `loadSoundfont` is called, all presets in the loaded soundfont are added to the available sounds of `s`.
Some type of sound registry has to be implemented to be able to add sounds from outside (webaudio package should not depend on soundfont package). Also remove the old non sf2 file based soundfonts

felixroos:
idea: add a new function that takes in a callback for creating the source web audio node:

```js
note("c a f e").source((hap, deadline, duration, cps) => {
  /* webaudio code that uses above params to create a webaudio node * /
  return node;
}).lpf(1000)
```

the function is similar to onTrigger, but instead of being just a general purpose callback, the function is expected to return a web audio node, which is used by the webaudio output to apply the standard chain. Advantages:

- allows to add custom sound sources from the outside. the existing oscillators, the sampler and soundfonts could all just implement a source.
- the standard chain is preserved, meaning a custom source will still work with all the effects (like lpf in the example)
- it can also be used in the user code to create custom instruments with webaudio code

felixroos:
this works really well! it opens up some interesting possibilities:

```js
const lfo = (freq, type = 'sine', range = 1) => {
  const ctx = getAudioContext();
  const osc = ctx.createOscillator();
  osc.type = type;
  osc.frequency.value = freq;
  osc.start();
  const g = new GainNode(ctx, { gain: range/2 });
  osc.connect(g); // -range, range
  return g;
}
const pitchLFO = lfo(.1, 'sine', 20);

note("c eb a*4 c4".add("12,12.05")).legato(.5 )
  .source(({ hap, t, duration }) => {
    const { release = 0.01 } = hap.value;
    const ctx = getAudioContext();
    const o = ctx.createOscillator();
    o.type = 'sawtooth';
    const freq = getFrequency(hap);
    o.frequency.value = Number(freq/2);
    o.start(t);
    o.stop(t + duration + release)
    pitchLFO.connect(o.frequency)
    return o;
  })
  .room(.5).lpf(800)
  .gain(.5)
```

this cries for a thin web audio abstraction layer..

felixroos:
PR for sound registry https://github.com/tidalcycles/strudel/pull/516

felixroos:
[...]

PR #435, opened 2023-02-14
WAM2 output

boourns:
This adds basic WAM2 loading, playing, GUI rendering to Strudel.

Definitely TODO, from my perspective:
- [ ] `.wam('inst').param('xxx', 0.1)` isn't working yet.  When a `param` is chained onto the wam, the instrument gets defaulted back to the default tone instrument and the wam is not played.
- [ ] WAM patch state is not saved in the URL.  We should call `instance.getState()` to pull the entire patch state

Would be nice:
- [ ] WAM GUIs should be cleaned up with `instance.destroyGUI(element)` as some GUIs otherwise hold onto a reference.

Future work:
- integrating into the same webaudio chain as the other instruments, so we can chain strudel effects after a WAM.  To me this would be the sensible point to also add support for inserting WAM audio effects into the effect chain.
- I liked @felixroos ideas mentioned in Discord around WAMs and samples being interchangeable so you could cycle through different options, but I think it's beyond my capabilities with the existing code and we could probably figure it out faster collaboratively.

### DIFF
--- a/index.mjs
+++ b/index.mjs
@@ -15,3 +15,4 @@ export * from './packages/transpiler/index.mjs';
 export * from './packages/webaudio/index.mjs';
 export * from './packages/webdirt/index.mjs';
 export * from './packages/xen/index.mjs';
+export * from './packages/wam/index.mjs';
--- a/packages/wam/index.mjs
+++ b/packages/wam/index.mjs
@@ -6,7 +6,8 @@ import {initializeWamHost} from '@webaudiomodules/sdk'
 let wams = {};
 
 // this is a map of all WebAudioModule instances (possibly more than one per WAM)
-let instances = {};
+let wamInstances = {};
+export const getWamInstances = () => wamInstances;
 
 // has the WAM host been initialized?
 let initialized = false
@@ -20,8 +21,8 @@ export const loadWAM = async function (name, url, what) {
         initialized = true
     }
 
-    if (!!instances[name]) {
-        return instances[name]
+    if (!!wamInstances[name]) {
+        return wamInstances[name]
     }
     
     if (!wams[url]) {
@@ -38,7 +39,7 @@ export const loadWAM = async function (name, url, what) {
 
     instance.audioNode.connect(getAudioContext().destination);
 
-    instances[name] = instance
+    wamInstances[name] = instance
 
     return instance
 }
@@ -48,7 +49,7 @@ export const loadWam = loadWAM;
 
 export const wam = register('wam', function (name, pat) {
     return pat.onTrigger((time, hap) => {
-        let i = instances[name]
+        let i = wamInstances[name]
 
         if (!i) {
             return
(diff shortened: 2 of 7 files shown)


felixroos:
thank you for this great contribution! I just dug a little bit in the code and found a way to implement `.param` . Also, I made some tweaks to the initWAM function to memoize promises instead of results (it could happen that the user reevaluates while a promise is pending). Example:

```js
await loadWAM('sh', 'https://editor.sequencer.party/plugins/burns-audio/synth101/index.js');

note("c3 c4 g3 eb3")
  .wam('sh')
  .param('filterFreq', "0.2 0.8 .1")
```

boourns:
Excellent! I will pull these changes and see how it runs.

felixroos:
> I think it would make sense to fix saving the WAM patch state or provide a way to easily set a preset.

I wonder how to handle the params that are dynamically controlled, which change in real time.. maybe those are just ignored?

> add the WAM state to the URL

Currently, the url is updated after a successful evaluation, containing the user code as base64.
The question is when to persist the wam state to the url? 

- either in real time (could be a bit heavy for drag events) 
- or when the code is reevaluated (not heavy, but could lead to people forgetting to evaluate before sharing).

I am generally a bit reluctant when it comes to adding additional state besides the code. I like the idea of having the whole state represented in the user code, to have one source of truth. Having two states would also complicate the api of the strudel package..
Maybe there is a way to actually write the ui state back to the code.. something like

```js
wamset('sh', {
  lfoRate: .5,
  pulseWidth: .6
})
```

this function would grab the current instance `sh` and make those changes on evaluation. 
All params that are untouched could remain undefined. When the user changes a param in the ui, 
the corresponding property is updated in the code (or added if it does not exist yet). 
The other way around already works with wam modules (ui changes when updated programmatically)

While this might blow up the user code a little, it will automatically work with the current url state handling.
And when strudel is used as a library, the object can be used as a text interface without ui too.

Codemirror also has [replacement decorations](https://codemirror.net/docs/ref/#view.Decoration%5Ereplace) which could be used to replace the `wamset` call with the wam gui in source (inside the code scrolling area)..

boourns:
Later this evening I will write some WAM-interfacing code to build the minimal "preset" object, and then we can look at options for injecting that object into the code.

I think ignoring dynamically set parameters makes the most sense - we can build a list of dynamically modified parameters in the 'param' method, although it would be good to reset that list every time the code is updated in case the user removes an automation.

What I do in sequencer party is periodically check parameter state for any WAM that has its GUI on screen, in a setInterval.  We could do something similar for whatever WAM is displayed in the footer.

boourns:
Added a function to build the minimal 'preset' object.  It's just wired up to a button labelled 'preset' on the WAM page for now.

I think it can be rewritten to save fewer variables for the loaded WAM but am not going to be working on this any more today or likely tomorrow so thought I'd share what I've got.

- [ ] automated parameters map is not cleared when code is updated, not sure how to do that
- [ ] WAM GUI cleanup doesn't occur when changing tabs and the GUI doesn't get reused when re-selecting the tab, so that should get cleaned up properly

felixroos:
this might be the way to pipe the wam node through the standard effects: https://github.com/tidalcycles/strudel/issues/182#issuecomment-1457100598 (not merged yet)

PR #461, opened 2023-02-16
FIXES: Warning about jsxBracketSameLine deprecation

bwagner:
See https://prettier.io/blog/2021/09/09/2.4.0.html

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -7,7 +7,7 @@
   "jsxSingleQuote": false,
   "trailingComma": "all",
   "bracketSpacing": true,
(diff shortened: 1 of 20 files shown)


bwagner:
See https://prettier.io/blog/2021/09/09/2.4.0.html

felixroos:
oh great! that adds a good guiding thread for readers. thanks :)

bwagner:
> oh great! that adds a good guiding thread for readers. thanks :)

I didn't mean to squash the config with the thread. Have to figure out how to start a separate PR, when another it's still pending.

Issue #380, opened 2023-01-16
slice function


felixroos:
https://strudel.tidalcycles.org/?k_CZOrXs1A-n first arg not patternified yet.. did not work with register because outerJoin is needed instead of innerJoin (at least if the function should work like in tidal, but I like the way it is). maybe we need a way to define the default join / define the others automatically..

yaxu:
Oops forgot about this one! I think my solution in #466 of reordering the parameters works well in this case though.

felixroos:
hm for me it looks more logical to call slice on the s param like that:

```js
s('jz_gtr').slice(8, "[3@3 3@2]*2 2*[3 6] 1*4 0*2")
// or
slice(8, "[3@3 3@2]*2 2*[3 6] 1*4 0*2", s('jz_gtr'))
```

isn't this also how tidal does it? With #466 it looks like:

```js
"[3@3 3@2]*2 2*[3 6] 1*4 0*2".slice(8, s('jz_gtr'))
// or
slice(8, s('jz_gtr'), "[3@3 3@2]*2 2*[3 6] 1*4 0*2")
```

in the latter case, both notations require 2 level nesting, while the former case only requires 2 level nesting in the second form.

yaxu:
You can miss off the s() to remove the nesting:
```js
"[3@3 3@2]*2 2*[3 6] 1*4 0*2".slice(8, 'jz_gtr')
// or
slice(8, "jz_gtr", "[3@3 3@2]*2 2*[3 6] 1*4 0*2")
```
This is consistent with the other functions then - where the most important thing is structure, and so it comes last or is the pattern the method is called on.

But  yep tidal does it the other way which I'm more used to, maybe it does read nicer to start with the control pattern..

yaxu:
Ok I think this is solved here as described above:
  https://github.com/tidalcycles/strudel/pull/466/commits/4918fbfaaf1330ae7eac8799b09935a77cb81c63

It was more convenient to use the monadic bind rather than applicative.. I'm not fully convinced that I got the right mix of inner and outer binds but it seems to check out.

yaxu:
We could still implement the alternative order of parameters, called e.g. `slicing` and `splicing`, if there is demand.

yaxu:
Implemented in #466

PR #466, opened 2023-02-18
slice and splice

yaxu:
Needs docs and tests


felixroos:
ref https://github.com/tidalcycles/strudel/issues/380

PR #467, opened 2023-02-19
settings tab with vim / emacs modes + additional themes and fonts

felixroos:
renamed theme tab to settings. themes can now be picked from a dropdown + added a checkbox to toggle vim mode
settings are now set up in a way that it is easy to add new settings + everything is persisted in the local storage

new themes:

image

using some fonts from https://github.com/ryanoasis/nerd-fonts/tree/master/src/unpatched-fonts

new settings tab:

image

update: the settings now also track the active footer tab, zen mode and latest evaluated code!


felixroos:
new themes:

image

using some fonts from https://github.com/ryanoasis/nerd-fonts/tree/master/src/unpatched-fonts

PR #476, opened 2023-02-23
add gainpre + gainpost for effect sends

felixroos:
to control the gain of the signal before and after effect sends. as discovered by @bwagner , it was not possible to get only the effect output without the original signal ( fully wet ).


felixroos:
additionally, there could be `delaywet` and `roomwet`. for example:

```js
s("bd")
.delaywet(0) // .delay(0)
.delaywet(.25) //delay(.5)
.delaywet(0.5) // .delay(1)
.delaywet(0.75) // delay(1).postgain(.5)
.delaywet(1) // delay(1).postgain(0)
```

- the basic idea is to crossfade between dry wet, where 0 - 0.5 dials in the effect and 0.5 - 1 dials out the signal
- in the current state, `postgain` affects both delay and room. ideally they should be decoupled.

Issue #192, opened 2022-08-16
Flesh out midi()


yaxu:
Relates to #60

Wish list:

```js
ccn(a).ccv(b).midi() // Sends cc a with value b
nrpnn(a).nrpv(b).midi() // Sends nrprn a with value b
midinote("40 20").midi() // sends midi notes
n(a).midi() // send midi note a.add(60) 
n(5).ccn(20).ccv(40).midi() // send a note and a cc in one pattern
```
The above send to first midi channel, send to both second and sixth like:
```js
midinote("40 20").channel("[1,5]").midi()
```
Also `midinote`, `polytouch`, `midibend`, etc..

felixroos:
nice wish list! would also love to find a way to support mpe for microtonal stuff

yaxu:
I think these mostly would work already via superdirt e.g. `ccn(30).ccv(70).superdirt()`

Tidal also has grouped parameters for cc and nrpn so this is possible `cc("30:70")`, it'd be nice to support that here too!

felixroos:
started implementing this in https://github.com/tidalcycles/strudel/issues/192 there is this list of controls: https://webmidijs.org/api/classes/OutputChannel#sendControlChange

clickglue:
Hi, thank you so much for all the work you put in Strudel, I really love playing it with digitakt/tone! Is there already a way to sync the midi clock?

felixroos:
Strudel currently does not support clock syncing, but I think the library it uses supports clocks: https://webmidijs.org/api/classes/Input#event:clock / https://webmidijs.org/api/classes/Output#sendClock I think strudel as the clock source would be simpler to implement, because the scheduler would not need syncing

yaxu:
Yes it would be a matter of calling `.sendClock` 48 times per cycle (assuming one 'note' per cycle), and `.sendClock`once every 4 cycles or so. In tidal this looks like `midicmd "[midiClock*48, start/4]" # s "midi"`. It could be nice to abstract this a bit to make it friendlier but it's quite nice to be able to tweak the numbers directly.

Here's superdirt's logic for turning messages into midi:
  https://github.com/musikinformatik/SuperDirt/blob/develop/classes/DirtEventTypes.sc#L40

There's some nuances e.g. if a control change arrives in the same message as a note you want to send the control change first. (This can delay notes a bit though.. midi gets fiddly)

felixroos:
[...]

PR #478, opened 2023-02-25
midi cc support

felixroos:
- now supports ccn and ccv
- also added midichan to select midi channel
- .midi now only chooses the output


felixroos:
ref https://github.com/tidalcycles/strudel/issues/192

PR #484, opened 2023-02-27
Revert "Another attempt at composable functions - WIP (#390)"

felixroos:
I noticed some heavy performance issues lately, so I dug around in the commits and found the change to be this commit.
Something must be in there that triggers a lot of "TOO LATE" messages, also sometimes breaking the whole main thread..
I think we should publish this now and investigate what the reason is.

I had to comment out slice, splice, weave and weaveWith as they relied on some changes in that commit.


yaxu:
Ok!

Issue #470, opened 2023-02-22
Global clock tempo


yaxu:
Related to #51 but a more general issue.

Without a global clock we're kind of stuck at 1 cps.

Slowing a pattern down works to some extent, but breaks functions that assume a global tempo like `splice`. Changing or patterning that also causes discontinuities.

It's be nice to have cps visible in the UI too, and maybe directly editable there.

PR #493, opened 2023-02-28
implement cps in scheduler

felixroos:
partly fixes https://github.com/tidalcycles/strudel/issues/470

- refactored the scheduler to use cps
- cps is now passed to hap onTrigger
- added setcps / setCps methods

still missing:

- patternable cps: how should this work? should the tempo have an effect on the cps pattern too?
- ui controls: basic version already implemented, but now commented out because it's not clear how the tempo is shared, as it is not part of the code

I'd say having the basic cps functionality working is already a good first step, the missing things can be addressed in a future PR

### DIFF
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -10,28 +10,35 @@ import { logger } from './logger.mjs';
 export class Cyclist {
   constructor({ interval, onTrigger, onToggle, onError, getTime, latency = 0.1 }) {
     this.started = false;
-    this.cps = 1; // TODO
-    this.phase = 0;
-    this.getTime = getTime;
+    this.cps = 1;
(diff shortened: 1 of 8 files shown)


yaxu:
oh great !!

> * patternable cps: how should this work? should the tempo have an effect on the cps pattern too?

In tidal it's just another control, so you could do e.g. `s("bd sd").cps("<0.75 1.5>")` or separately `cps(saw.range(1,2).slow(4).segment(32))`.

This gets a bit complicated for the scheduler. If it's running at 20Hz, it queries the pattern for the right slice, but has to deal with changes in cps during that slice which means the slice is then either too big or too small..

In tidal `setcps` is just an alias for `once . cps`

> I'd say having the basic cps functionality working is already a good first step, the missing things can be addressed in a future PR

agreed!

Issue #499, opened 2023-03-01
looping samples


felixroos:
there is already `loop`, but it could be extended with `loopStart` / `loopEnd` to set a loop the sample runs into.
this is nice for creating sustained notes from samples, like soundfonts do. the implementation should be pretty easy, as webaudio already supports loopStart and loopEnd. probably need to respect `begin` and `end`...

yaxu:
`loopBegin` might fit better with `begin`

felixroos:
fixed with: https://github.com/tidalcycles/strudel/pull/698

PR #503, opened 2023-03-02
add loopBegin and loopEnd

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/499

As soon as one of these is defined, the sample will loop from loopBegin || 0 to loopEnd || 1

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -193,16 +193,34 @@ const generic_params = [
    */
   ['end'],
   /**
-   * Loops the sample (from `begin` to `end`) the specified number of times.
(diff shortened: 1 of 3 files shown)


felixroos:
should also be interesting to use soundfont buffers as samples and set the loop points exactly to the correct location to get basically soundfonts without soundfonts. here is a first attempt to convert soundfont buffers to wav files: https://stackblitz.com/edit/js-6yzedj?file=index.js,index.html,sample.js

felixroos:
superseded by https://github.com/tidalcycles/strudel/pull/698

PR #513, opened 2023-03-05
Memories WIP

yaxu:


yaxu:
This mirrors Tidal's use of `^` for embedding patterns, but strudel uses `^` in some chords. We could support both but maybe better to standardise?

felixroos:
just realized i have never answered this... looks like a good thing to have. so this basically adds variable support to the mini notation?

phylonyus:
`$` prefix would match lots of other programming languages for string interpolation, e.g. this:
```javascript
remember('foo', "0 1 2")
remember('bar', "0 2 1")
"$foo <$foo $bar>".scale("C2:major).n()
```
would resolve to cycles: `"0 1 2 0 1 2", "0 1 2 0 2 1"`

can this use var/let instead, and the repl parser registers the Pattern for me into the mini parser? e.g.:
```javascript
let foo = "x(3,8)"
s("bd").struct("$foo [x x]")
```

yaxu:
This is pattern interpolation rather than string interpolation, and the `^` syntax is already established in tidal.

It's also reactive rather than being scoped `let` style assignment. I.e. in

```js
remember('nice', "1 2 3")
let x = "^nice"
remember('nice', "4 5 6")
```

`x` would take on the new value of `nice`. This would become important in a conversational strudel interface.

Issue #268, opened 2022-11-22
Reprint the "samples" tab when samples are added.


mr-eggleton:
When samples() is called :

```javascript
// Load the 'default samples' from tidal/superdirt
await samples('github:tidalcycles/Dirt-Samples/master')
```

Reprint the samples list in the tab (Currently starts "826 banks loaded:") to contain the new things you have loaded.

felixroos:
good point.. lost track on that one! the samples tab is pretty underdeveloped atm

mr-eggleton:
Was in Alex's workshop in Sheffield tonight and l noticed it pretty quickly. I'm a FoxDot user and we have a similar feature that I use all the time.

Well done on Strudel it's a great price of work and developing so fast. The kids in my Algorave club are looking forward to learning better.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

felixroos:
fixed in https://github.com/tidalcycles/strudel/pull/516 !

felixroos:
the "samples" tab is now called "sounds", listing all sounds you can use with s , grouped into "samples", "drum-machines", "synths" and "user". It will also update live if you're loading samples via samples. Inside synths, there is now a large selection of general midi soundfonts, so you can finally write music with helicopters and gunshots 🔫 here is a demo using soundfonts: https://strudel.tidalcycles.org/?PYLKXJgtX9rV another cool thing you can do now is defining custom web audio nodes via source for full custom synthesis using the web audio api, demo: https://strudel.tidalcycles.org/?ymSQTC8hc9WC these changing involved refactoring parts of the webaudio package, so there could be some bugs floating around.. if you find one, please shout 📣  there should generally be no breaking changes (except the old, nowhere documented soundfont names do not work anymore..)

PR #516, opened 2023-03-05
registerSound API + improved sounds tab + regroup soundfonts

felixroos:
- [x] refactor sampler to use it
- [x] refactor synth to use it
- [x] add 'source' control
- [x] overhaul sounds tab
- [x] soundfonts
- [x] bug: cut sometimes fails to cut the stop the event
- [x] ~~migrate webdirt?~~


felixroos:
fixes https://github.com/tidalcycles/strudel/issues/268

felixroos:
this should be ready now.. i refactored the way sounds are registered. it's similar to how pattern methods are registered, using a map to add sounds from outside. I documented the details [in sounds.mdx](https://github.com/tidalcycles/strudel/pull/516/files#diff-8efbb883ae4293b7be49241e995e730d397f3c25058dfa3441df39eeae1aacda).
the `samples` function now also uses this internally.

Most of this PR is implementation details, but some things are also noticeable on the surface: 

- that the samples tab is now called sounds, listing all registered sounds in real time, which fixes https://github.com/tidalcycles/strudel/issues/268 .
- I generally improved the sounds tab a bit, grouping sounds into "samples", "drum-machines", "synths", and "user".
- Soundfonts are now regrouped and curated (threw out the buggy ones) to use meaningful names!
- Added a new param "source" that is given a callback to output a custom webaudio node. This could be the beginning of a more flexible approach to synthesis in user land. Will post an example of this soon

PR #521, opened 2023-03-17
fix: envelopes in chrome

felixroos:
as reported in by @bwagner , chrome envelopes behave differently.. The reason is "cancelAndHoldAtTime" seems to behave differently than I thought. I commented it out for now.. envelopes in webaudio seem to be a bit messed up :-/
Further investigation is needed


felixroos:
follow up issue: https://github.com/tidalcycles/strudel/issues/522

PR #527, opened 2023-03-18
FIXES: alias pm for polymeter

bwagner:
(return statement was missing)

I was looking at [polymeter](https://strudel.tidalcycles.org/learn/factories#polymeter) and when replacing `polymeter` by `pm` in the MiniRepl, I got error `"Cannot read properties of undefined (reading 'note')".`

FIXES: Issue https://github.com/tidalcycles/strudel/issues/532

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -1127,6 +1127,8 @@ function _composeOp(a, b, func) {
 export const polyrhythm = stack;
 export const pr = stack;
 
+export const pm = polymeter;
+
 // methods that create patterns, which are added to patternified Pattern methods
 // TODO: remove? this is only used in old transpiler (shapeshifter)
 Pattern.prototype.factories = {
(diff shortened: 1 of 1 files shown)


packages/core/pattern.mjs
```diff
@@ -1376,7 +1376,7 @@ export function polymeter(...args) {
 
 // alias
 export function pm(...args) {
```
yaxu, line 1378: Aha thanks for the fix!

While you're here, it would probably be better to just have `export pm = polymeter;` instead of the function

yaxu (changes requested):

packages/core/pattern.mjs
bwagner: I had tried that before, but it causes an error at startup:
```
pattern.mjs?t=1679428469545:1142 Uncaught (in promise) ReferenceError: Cannot access 'pm' before initialization
    at pattern.mjs?t=1679428469545:1142:3
```
This error stems from this code:

```
// methods that create patterns, which are added to patternified Pattern methods
// TODO: remove? this is only used in old transpiler (shapeshifter)
Pattern.prototype.factories = {
  pure,
  stack,
  slowcat,
  fastcat,
  cat,
  timeCat,
  sequence,
  seq,
  polymeter,
  pm,
  polyrhythm,
  pr,
};
// the magic happens in Pattern constructor. Keeping this in prototype enables adding methods from the outside (e.g. see tonal.ts)

```

So I shifted `export pm = polymeter;` before that `Pattern.prototype.factories` statement.

PR #529, opened 2023-03-20
fix(footer): fix link to tidalcycles

revolunet:
Hi ! thanks a ton for strudle, amazing thing !


felixroos:
thanks for the fix :)

PR #295, opened 2022-12-10
Move stuff to new register function

felixroos:
getting rid of all usages of `Pattern.prototype.define`,`Pattern.prototype.bootstrap`, `Pattern.patternify`, `patternifyN`

PR #531, opened 2023-03-20
Maintain random seed state in parser, not globally

### LINKED ISSUE #530: `rand` produces duplicate values when used repeatedly in `seq()`
Hello! I've observed a funny interaction between `rand` and `seq()`.

I initially noticed that `s("[sd|cp] [sd|cp]")` will always select both `sd`s or both `cp`s in any given cycle; you never get the combinations `sd cp` or `cp sd`. This is surprising because it seems like the choices should be independent.

I confirmed that this is different than Tidal's behavior: after evaluating `d1 $ s "bd [sd|cp] [sd|cp]"`, I observe all four combinations of [sd, cp] x [sd, cp].

The `|` syntax is mini-notation for `chooseCycles()`, which uses `rand`; I can see that `rand` itself is producing duplicate values: 
```js
s("hh*2").gain(seq(rand, rand).log())
```
generates console output like
```
[hap] 136/1 → 273/2: s:hh gain:0.13287117518484592
[hap] 273/2 → 137/1: s:hh gain:0.13287117518484592
[hap] 137/1 → 275/2: s:hh gain:0.1258764062076807
[hap] 275/2 → 138/1: s:hh gain:0.1258764062076807
[hap] 138/1 → 277/2: s:hh gain:0.564989872276783
[hap] 277/2 → 139/1: s:hh gain:0.564989872276783
```
However, this does not occur with `s("hh*2").gain(rand.log())`.

(`s("hh").gain(rand.log()).pan(rand.log())` also produces duplicate values for gain and pan, but I get the impression that's expected behavior because the PRNG is a pure function of time, and the two uses of `rand` here are simultaneous for each hap.)

In short, this seems like a regression from the behavior described here: https://github.com/tidalcycles/strudel/pull/165#issuecomment-1200476739.

That's as far as I've gotten for now; I plan to take a closer look at this soon and hopefully submit a fix.

ijc8:
Fixes #530.

An alternative approach would be to defer this until `patternifyAST()`, more like in @bpow's [original version](https://github.com/bpow/strudel/blob/984f3d2781a8b8ad2c96ffa9dfb56bf431eec918/packages/mini/mini.mjs), in which seeds are generated at AST->pattern conversion time. I think this would work fine if the state was reset for each top-level call to `patternifyAST()`, which would also effectively result in per-parse state.
The approach I've taken here is more like how Tidal does it, where the seeds are generated at parse time.

### DIFF
--- a/packages/mini/krill-parser.js
+++ b/packages/mini/krill-parser.js
@@ -279,7 +279,7 @@ function peg$parse(input, options) {
   var peg$f8 = function(p, s, r) { return x => x.options_['ops'].push({ type_: "bjorklund", arguments_ :{ pulse: p, step:s, rotation:r }}) };
   var peg$f9 = function(a) { return x => x.options_['ops'].push({ type_: "stretch", arguments_ :{ amount:a, type: 'slow' }}) };
   var peg$f10 = function(a) { return x => x.options_['ops'].push({ type_: "stretch", arguments_ :{ amount:a, type: 'fast' }}) };
-  var peg$f11 = function(a) { return x => x.options_['ops'].push({ type_: "degradeBy", arguments_ :{ amount:a } }) };
+  var peg$f11 = function(a) { return x => x.options_['ops'].push({ type_: "degradeBy", arguments_ :{ amount:a, seed: seed++ } }) };
   var peg$f12 = function(s) { return x => x.options_['ops'].push({ type_: "tail", arguments_ :{ element:s } }) };
   var peg$f13 = function(s, ops) { const result = new ElementStub(s, {ops: [], weight: 1, reps: 1});
     for (const op of ops) {
@@ -289,8 +289,8 @@ function peg$parse(input, options) {
   };
   var peg$f14 = function(s) { return new PatternStub(s, 'fastcat'); };
   var peg$f15 = function(tail) { return { alignment: 'stack', list: tail }; };
-  var peg$f16 = function(tail) { return { alignment: 'rand', list: tail }; };
-  var peg$f17 = function(head, tail) { if (tail && tail.list.length > 0) { return new PatternStub([head, ...tail.list], tail.alignment); } else { return head; } };
+  var peg$f16 = function(tail) { return { alignment: 'rand', list: tail, seed: seed++ }; };
+  var peg$f17 = function(head, tail) { if (tail && tail.list.length > 0) { return new PatternStub([head, ...tail.list], tail.alignment, tail.seed); } else { return head; } };
   var peg$f18 = function(head, tail) { return new PatternStub(tail ? [head, ...tail.list] : [head], 'polymeter'); };
   var peg$f19 = function(sc) { return sc; };
   var peg$f20 = function(s) { return { name: "struct", args: { mini:s }}};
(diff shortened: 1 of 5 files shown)


ijc8:
Quick tests:
`s("[sd|cp] [sd|cp]")` - previously made the same choice for both (`sd sd` or `cp cp`), now all four combinations occur
`s("bd? sd?")` - previously played both or neither; now can play one or the other
`"[c3|e3|a3],[c4|e4|a4]".note()` (from https://github.com/tidalcycles/strudel/pull/165#issuecomment-1200476739) - previously played octaves, now plays various intervals

Also,
```js
stack(
  s("hh*8").degrade(),
  s("[ht*8]?")
)
```
is still in sync, which is also the case on `main` (but presumably wasn't between #165 and #295).

Evaluating `console.log(s("bd? sd?").queryArc(0, 1).map(e => e.show()))` repeatedly always yields the same result, as parsing & pattern construction are free of side-effects (because random seeds are only incremented within each parse, not between parses).

ijc8:
I tried writing a new test to confirm that all possibilities occur with equal probability... only to find that it's not the case!

Here's my experiment:
```js
const numCycles = 1000;
const values = mini('[a|b] [a|b]').queryArc(0, numCycles).map(e => e.value);
const observed = { 'aa': 0, 'ab': 0, 'ba': 0, 'bb': 0 }
for (let i = 0; i < values.length; i += 2) {
    const chunk = values.slice(i, i + 2);
    observed[chunk.join('')]++;
}
console.log(observed);
```
Running on `main`, I observe `{ aa: 502, ab: 0, ba: 0, bb: 498 }`; as expected, the alternating pairs never happen.

However, on my branch, I don't observe an even spread amongst the four options as I'd expect, but instead `{ aa: 344, ab: 158, ba: 151, bb: 347 }`. All four combinations now occur, but there is still a strong bias towards the non-alternating combinations.

I see that `rand` only has [a period of 300 cycles](https://github.com/tidalcycles/strudel/blob/952bd38e70759dde359d9307a8bade52a76ce30d/packages/core/signal.mjs#L93), so I suppose it's only the first 300 cycles that count for this; with `numCycles = 300`, I get `{ aa: 104, ab: 46, ba: 46, bb: 104 }`.

For comparison, in Tidal the distribution is more even:
```hs
import Data.Maybe
import Data.List.Split
import qualified Data.Map as Map

pairs = chunksOf 2 $ concat $ map eventValue $ queryArc ("[a|b] [a|b]" :: Pattern String) (Arc 0 300)

foldr (\k m -> Map.alter (\n -> Just $ fromMaybe 0 n + 1) k m) Map.empty pairs
```
the last expression evaluates to
```haskell
fromList [("aa",80),("ab",70),("ba",70),("bb",80)]
```

I wondered if the difference was due to `randOffset` (0.0001 in Tidal vs. 0.0002 in Strudel) and tried changing it. To my surprise, I found that the distribution was very sensitive to this value. At 0.0001 it's even worse (`{ aa: 126, ab: 24, ba: 24, bb: 126 }`) and at 0.0003 it's better (`{ aa: 78, ab: 72, ba: 72, bb: 78 }`).

For the time being, I'll just change `randOffset` to 0.0003 and add the new test.

felixroos:
great to see you've managed to get it to work, thank you so much :) I am surprised that no tests failed... Probably because all test tunes (tunes.mjs) that use "|" or "?" are only tested for one cycle, and the first cycle is still the same.

In the future, we could think about  if it makes sense to move that seed state from the mini parser to the randomness functions themselves, as using `chooseInWith` or `degradeByWith` without mini notation is still showing the stateless behavior.

But for now, I'd say this could be merged as is, as it makes randomness much more usable!

PR #535, opened 2023-03-23
Update lerna

felixroos:
just tested how batch publishing works now with pnpm. I've updated lerna to work with pnpm.
see contributing.md for info on how to publish. basically, lerna is now only used for version bumping, while pnpm does the publishing. maybe it makes sense to replace lerna with https://github.com/changesets/changesets at some point, but it's fine for now

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,7 +123,14 @@ To publish all packages that have been changed since the last release, run:
 
 ```sh
 npm login
-npx lerna publish
+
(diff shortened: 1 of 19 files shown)


felixroos:
btw the 0.7.0 versions are broken, because I had to find out that lerna does not use pnpm for publishing, which causes it to not apply the publishConfig as pnpm does, which breaks the imports (using unbuilt esm files instead of built esm / cjs)... everything is now working on either 0.7.1 or 0.7.2

PR #557, opened 2023-05-02
add basic triads and guidetone voicings

felixroos:
should have been there some time ago..


felixroos:
example with guidetones: https://strudel.tidalcycles.org/?J6W1EjvZmiN2 <3

PR #563, opened 2023-05-05
Vanilla JS Refactoring

felixroos:
implement some things in vanilla js that were previously only in react land:

- more sophisticated `Drawer` to schedule drawing haps in past present and future, similar to `usePatternFrame` hook
- new `@strudel/codemirror` package that uses `Drawer` + repl as part of new `StrudelMirror` interface
- added a new example "vite-vanilla-repl-cm6" as a demo
- removed time.mjs + refactored usages
- new `@strudel/web` package that simplifies using strudel as a library

TODO:

- [x] integrate relevant bits into core package
- [x] create codemirror package

in a future PR, the react package can also be refactored to use `Drawer` (and maybe codemirror package?) (https://github.com/tidalcycles/strudel/issues/564)
I am also thinking about reimplementing the main repl in SolidJS :P

also check out [this sandbox](https://codesandbox.io/s/vanilla-codemirror-strudel-forked-35g7wg?file=/src/index.js)


felixroos:
finally merging this now..

PR #570, opened 2023-05-24
Solmization added

dariacotocu:
Added alternative solmization types for the notes. I added to the code : solfege, german, indian, byzantine and japanese.
Also created a file in order to test if it works with the terminal.


dariacotocu:
i think the test file should work now! i made it into vitest form. 
Yes, that is a good point regarding the control function, it would make it clearer. I will have a look at how to do it :)

felixroos:
I am just reading about indian note systems. there is a system of 12 notes called svaras: https://en.wikipedia.org/wiki/Svara they have slight regional differences. At first sight, the most friendly in terms of textual notation is the carnatic one: https://en.wikipedia.org/wiki/Svara#Svaras_in_Carnatic_music (see Mnemonic). The Hindustani system normally has underlines to denote flats. In some corners of the internet they are notated uppercase (major scale) and lowercase (pentatonic scale): Sa re Re ga Ga Ma ma? Pa dha Dha ni Ni. You also see only single letters sometimes: S r R g G M m? P d D n N. (Not sure about the sharp fourth). It probably makes sense to ask actual indian classical musicians about that.. It also does not seem to be a 1:1 mapping for 12 equal divisions of the octave...

felixroos:
@dariacotocu do you want to develop this further or should we merge the midi2note function as is to be developed further in a later PR?

dariacotocu:
i can try looking into it further, but i would not mind you merging the midi2note function to be developed further! when i get some free time i would love to look into the issue again and think about it more in-depth :) thank you

felixroos:
> i can try looking into it further, but i would not mind you merging the midi2note function to be developed further! 

ok, I'll merge it now, otherwise it'll get too many diffs later.. There was already a little conflict, which should now be fixed (I moved your logic to a new function called sol2note).

> when i get some free time i would love to look into the issue again and think about it more in-depth :) thank you

thank you!

Issue #574, opened 2023-06-02
UX proposal: repl sidebar


roipoussiere:
I find the bottom bar a bit disturbing. Among other things, it adds many scroll bars in the UI (see image below), an extra menu bar, and takes some space.

I wonder if it can be removed. Here are some suggestions to achieve this:

- `welcome`, `reference` and `sound` could be displayed as full-screen windows appearing when clicking on corresponding items in a new **help** menu located in the top bar (and so the current *learn* button could be part of it);
- `console` could be en/disabled by moving a slider (so we can also set its size);
- `settings` could be displayed as a full-screen window when clicking on a new **settings** button in the top bar.

I can try to work on it if you think it's a good idea.

---

![image](https://github.com/tidalcycles/strudel/assets/1665542/46636f03-93b6-4ae1-ae4f-f9c8e07feeaa)

felixroos:
> it adds many scroll bars in the UI

I'd say that's more a css bug.. I haven't noticed it yet, likely because my browsers / systems use floating scrollbars, so I only see the scrollbar that's actively used. So at least the second scrollbar could be removed by tweaking the style.

> and takes some space

you can also click the strudel icon to enter "zen mode" :) 

> I wonder if it can be removed

I kind of like the bottom bar. Especially for sounds and reference, imho it makes more sense to not have a full screen so you can quickly copy paste stuff around. It is also likely that the bottom bar will be extended / extendable in the future.

So maybe start by fixing the scrollbar? Maybe it's less disturbing already then? Would you want to try that?

We could maybe also add different UI variants under different routes, e.g. strudel.tidalcycles.org/xxx

roipoussiere:
> I kind of like the bottom bar. Especially for sounds and reference, imho it makes more sense to not have a full screen so you can quickly copy paste stuff around. It is also likely that the bottom bar will be extended / extendable in the future.

Ok I understand. When the bottom bar is open, there is a just a few lines of code displayed in the editor, so I can't easily use both the editor and the bottom bar content.

Also, the content of the current bottom bar seems to be better appropriate for a less wide container (see unused area bellow).

![image](https://github.com/tidalcycles/strudel/assets/1665542/214dfba5-99b1-4c4f-9c58-a93427e5dfb7)

So what about a side bar with editable width? We can have many lines of code displayed in the editor, while easily accessing the sounds, reference, etc. and have it bigger or smaller if necessary. He is a quick mock-up:

[...]

PR #581, opened 2023-06-06
editor: enable line wrapping

roipoussiere:
Useful when the editor has a small width (required for #574).


felixroos:
maybe also add a setting for that?

roipoussiere:
Yes I can, but I just wonder what is the use case of don't wanting line breaks?

I like to understand user needs before adding many options, which can overload the UI at some point.

felixroos:
> Yes I can, but I just wonder what is the use case of don't wanting line breaks?

- You could have a single line of code with a large amount of data, e.g. voicings
- You could actually want to scroll horizontally for layouting purposes, example: https://strudel.tidalcycles.org/?_usVdv9dq4rH

roipoussiere:
Ok thanks for the explications, I'm adding the option to the settings.

roipoussiere:
I'm waiting for #588 to be merged to avoid conflicts, as modifications takes in same places.

felixroos:
Just fixed the conflict

roipoussiere:
added the checkbox in last commit

roipoussiere:
It works pretty well with different screen sizes:

image

felixroos:
nice! so i suppose it's ready to merge?

roipoussiere:
yep!

PR #427, opened 2023-02-09
autocomplete preparations

felixroos:
uses jsdoc to populate autocomplete entries:

image

todo: be context aware, show methods only after a dot, show top level functions only on top level etc..
see https://codemirror.net/examples/autocompletion/ (Completing from Syntax)

PR #588, opened 2023-06-09
enable auto-completion

### LINKED ISSUE #426: repl: autocomplete
using https://codemirror.net/examples/autocompletion/ the editor should have an autocomplete that uses the data from jsdoc.

roipoussiere:
Now https://github.com/uiwjs/react-codemirror/pull/458 is merged, the auto-completion seems to work pretty well, so maybe we can enable it?
follows #427, fix #426

### DIFF
--- a/packages/react/src/components/CodeMirror6.jsx
+++ b/packages/react/src/components/CodeMirror6.jsx
@@ -133,20 +133,24 @@ export default function CodeMirror({
   );
 
   const extensions = useMemo(() => {
+    let _extensions = [...staticExtensions];
     let bindings = {
       vim,
       emacs,
     };
-    if (bindings[keybindings]) {
-      return [...staticExtensions, bindings[keybindings]()];
-    }
-    return staticExtensions;
-  }, [keybindings]);
 
-  const setAutoCompletion = (isAutoCompletionEnabled) =>
-    isAutoCompletionEnabled
-      ? javascriptLanguage.data.of({ autocomplete: strudelAutocomplete })
-      : autocompletion({ override: [] });
+    if (bindings[keybindings]) {
+      _extensions.push(bindings[keybindings]());
+    }
+
+    if (isAutoCompletionEnabled) {
+      _extensions.push(javascriptLanguage.data.of({ autocomplete: strudelAutocomplete }));
+    } else {
+      _extensions.push(autocompletion({ override: [] }));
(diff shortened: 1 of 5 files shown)


roipoussiere:
update: css fix to avoid this glitch:

image

yaxu:
It would be good to have an option for this, personally I'd prefer it default off.

felixroos:
thanks for reviving the autocomplete :) besides the now merged PR (https://github.com/uiwjs/react-codemirror/pull/458), there is one more thing still not done, as mentioned in https://github.com/tidalcycles/strudel/pull/427  :

> todo: be context aware, show methods only after a dot, show top level functions only on top level etc..
see https://codemirror.net/examples/autocompletion/ (Completing from Syntax)

without this, the autocomplete is pretty dumb, as it suggests any name at any place, so many things that get suggested don't actually work, for example:

```js
s("bd sd").s
```

this will suggest `samples`, which is not a pattern method, so it will throw an error. (I haven't tested this specific case but I remember that it did things like that).

maybe it makes sense to add this before merging? it could be a bit tricky to decide how to decide which goes what, and how to annotate it.
but we could also add it as an opt-in experimental feature as it is now..

roipoussiere:
I'm adding the button in the settings, so I can disable the feature by default until he problem you mention is not solved.

>  but we could also add it as an opt-in experimental feature as it is now..

roipoussiere:
> It would be good to have an option for this, personally I'd prefer it default off.

Done

packages/react/src/components/CodeMirror6.jsx
```diff
@@ -140,6 +142,12 @@ export default function CodeMirror({
     }
     return staticExtensions;
   }, [keybindings]);
+
+  const setAutoCompletion = (isAutoCompletionEnabled) =>
```
felixroos, line 146: why is this a function? couldn't this be added to extensions inside the useMemo (L135+) ?
outside of the memo the function will be created and the extensions will be spreaded on every render.
This creates a little bit of garbage on every keystroke

roipoussiere: Indeed, fixed

felixroos:
I'll merge this now with the current state, although it's currently sometimes suggesting things that are wrong, it's not on by default and already has value as-is.

Issue #288, opened 2022-12-09
simplify union of bare values


felixroos:
another value related thing that would be handy:

```js
stack(
"0 1 2",
"7 10 11"
).scale('E minor').note()
```

The above snippet works, but what if I want to add a cutoff to the first pattern?

```js
stack(
"0 1 2".cutoff(1000), // <---- this fails
"7 10 11"
).scale('E minor').note()
```

the only way to do this is to rewrite:

```js
stack(
"0 1 2".scale('E minor').note().cutoff(1000),
"7 10 11".scale('E minor').note()
)
```

which means a lot of duplication and rearranging.. the duplication **can** be solved by

```js
const h = x=>x.scale('E minor').note();
stack(
"0 1 2".layer(h).cutoff(1000),
"7 10 11".layer(h)
)
```

but that's not optimal.. It would be really convenient if the second snippet just worked.. here is it again:

```js
stack(
"0 1 2".cutoff(1000), // <---- this fails
"7 10 11"
).scale('E minor').note()
```

The problem: `"0 1 2".cutoff(1000)` generates values like `{ value:0, cutoff: 1000}`, which are then wrapped into note later, creating `{ note: { value:0, cutoff: 1000} }` . A possible solution would be to check if a control value is an object containing a value, and if thats the case, just use the value and merge the rest on top level, giving: `{ note: 0, cutoff: 1000}`. I don't see any drawbacks there, except that you cannot have objects as control values that contain the now protected "value" keyword.

felixroos:
and if that is really a problem, we could rename "value" to "__value" or something like that

yaxu:
Hmm I see the problem but find this solution a bit confusing, needs some more thought I think..

felixroos:
I don't see any alternative.. having this would make live coding so much easier, I run into this problem all the time.

yaxu:
One alternative would be:
```js
stack(
note("0 1 2").cutoff(3000),
note("7 10 11")
).scale('E minor')
```
Advantages:
- you don't have to search through the pattern to find out what the `"0 1 2"` is
- less of a conceptual leap as you don't have to imagine adding an effect to a bare value
- scales better as the pattern gets more complex, it's clear what everything is

Disadvantage:
- a bit more to type in

[...]

PR #598, opened 2023-06-12
clip now works like legato in tidal

felixroos:
- hap.duration now respects clip value
- hap.endClipped is now end*clip
- visualizations show clipped length
- clip(0) will now be silence

clip so far was only a flag. now it supports floats, making it equivalent to legato in tidal.
i did not touch legato because it would break dozens of existing patterns, as legato currently works without hap.value, so it can be used within a control param function.
in the long run, it should make sense to either refactor old patterns or implement https://github.com/tidalcycles/strudel/issues/288 . 
with just clip changed, the only breaking change is that clip(0) now brings silence, instead of, as currently, disabling clip.
In practice, clip(0) is not really in use, as it is the default behavior anyway (i haven't found a single use anywhere)

btw shoutout to henk from the workshop for inspiring the idea!


felixroos:
I have now refactored usages of `legato` in the examples to `clip` .. The question is if the old legato should stay as is (illegal time manipulation) or if we should change it to be an alias of clip (like tidal).. Changing it will break a lot of existing patterns, as the current legato can be used on non-object values. 
I'd say we merge this now so people won't see any more examples of a thing that might break in the future. We can decide later if we want to do the breaking change or implement https://github.com/tidalcycles/strudel/issues/288

ok?

felixroos:
@yaxu any objections?

yaxu:
Sorry for being slow, looks good!

yaxu:
> we should change it to be an alias of clip (like tidal)

actually clip/legato don't seem to be aliased in Tidal yet, will make an issue..

telephon:
Just to be sure this is really correct?
- if `legato` is given, it overrides the sample's own duration. It is the multiple of the inter-onset time (`delta`, received from tidalcycles). For `legato = 1`, one event ends where the next starts, for `legato = 0.5` it ends in between, for `legato = 2` it overlaps fully with the next event.
- if `sustain` is given, `legato` has no effect (sustain is the total duration of an event)

yaxu:
Yes I think that's the same as `clip`:
```js
note("c e f g").s("folkharp").clip("<0.5 1 2>")
```
https://strudel.cc/?ezJvywGYY2vq

It's not really inter-onset time though, but event duration. So `"a ~ b ~".clip(0.5)` translates to events of 1/8th of a cycle. I think is where the name 'legato' becomes problematic?

telephon:
> So `"a ~ b ~".clip(0.5)` translates to events of 1/8th of a cycle. I think is where the name 'legato' becomes problematic?

The inter-onset time for the events in the cycle is 1/4, so it makes sense that legato 0.5 would make them 1/8. What about clip on patterns with unequal duration? There, the "clipping" should be relative to their individual duration (actually: inter onset time).

PR #607, opened 2023-06-18
tonal fixes

felixroos:
Tonal API -> Tonal Functions
+ update scale syntax


felixroos:
also fixes https://github.com/tidalcycles/strudel/pull/605

PR #613, opened 2023-06-23
tauri test

vasilymilovidov:
After installing the dependancies, use 'pnpm tauri dev' to run or 'pnpm tauri build' to build.

I'm not that familiar with Strudel's codebase and have never built an app with Tauri before. This is just an attempt at following the official guide for Tauri integration. Everything seems to be working, except webmidi.

I tested it on macOS Ventura. There's some slight audio crackle from time to time, but I have the same issue in Safari.


felixroos:
amazing! and just ~18MB in size :) just tested it and it works!
would you mind adding a README.md to the src-tauri project, with a little guide on how to dev / build?
should also make sense to mention that you need to install rust, e.g. using https://rustup.rs/

I think webaudio is not so optimized on safari / webkit so it should make sense to put more effort into performance on webkit for macos and (some?) linux users (of course not in this PR). having a desktop app would also allow packaging a much faster audio engine.....

vasilymilovidov:
I was actually able to push it down to 15MB with some rust build optimization. Will push them and a README soon. 

I was also thinking about tinkering with github actions so it would build bundles and installers automatically for a bunch of platforms. What do you think?

felixroos:
> I was also thinking about tinkering with github actions so it would build bundles and installers automatically for a bunch of platforms. What do you think?

sounds good! probably best to build it on release.. found this on a brief research : https://github.com/softprops/action-gh-release

vasilymilovidov:
> sounds good! probably best to build it on release.. found this on a brief research : https://github.com/softprops/action-gh-release

I'll take a look. 
I'm also writing a midi sequencer in rust at the moment. I'll see if I can apply this knowledge to Strudel and make midi work from tauri  🙃

vasilymilovidov:
Added a separate workflow for tauri to not mess with existing ones. I think it should work.

felixroos:
great! so it will run automatically when a new version tag is pushed?

vasilymilovidov:
yep. tested it only on a dummy repo though. I think I'll make a private clone of strudel repo to test it properly.

vasilymilovidov:
tested it properly and made some adjustments. it builds for mac and windows, but there are issues with ubuntu.

vasilymilovidov:
everything should work now

felixroos:
so is it ready to merge then?

vasilymilovidov:
> so is it ready to merge then?

yes. I'll work on MIDI support next, but on a new branch.

roipoussiere:
Very nice to see this desktop app!

I just to mention that the .deb probably needs some description:

image

PR #617, opened 2023-06-28
standalone app midi WIP

vasilymilovidov:
Attempting to incorporate MIDI support for the standalone version of Strudel using the `midir` library in Rust.

What works:
— sending MIDI notes
— cc messages
— specifying channels
— changing MIDI ports
— legato
— velocity

What currently doesn't work right:
— durations and note-offs kinda work, but there are some glitches. needs better implementation 

TODOs:
— better note durations and note-offs
— improve velocity
— better message handling on the js side


packages/core/package.json
```diff
@@ -31,6 +31,7 @@
   },
   "homepage": "https://strudel.tidalcycles.org",
   "dependencies": {
+    "@tauri-apps/api": "^1.4.0",
```
felixroos, line 34: i think we should try to not have that dependency in the core module to keep it properly encapsulated

packages/midi/midi.mjs
```diff
@@ -66,8 +67,57 @@ function getDevice(output, outputs) {
...
+      const timingOffset = WebMidi.time - getAudioContext().currentTime * 1000;
+      time = time * 1000 + timingOffset;
+      // destructure value
+      const { note, nrpnn, nrpv, ccn, ccv, midichan = 1 } = hap.value;
+      const velocity = hap.context?.velocity ?? 0.9; // TODO: refactor velocity
+      const duration = hap.duration.valueOf() * 1000 - 100;
+
+      if (note) {
+        const midiNumber = noteToMidi(note);
+        setTimeout(() => {
```
felixroos, line 83: is there a way to avoid setTimeout? the timing of setTimeout is not to be trusted

packages/midi/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "homepage": "https://github.com/tidalcycles/strudel#readme",
   "dependencies": {
+    "@tauri-apps/api": "^1.4.0",
```
felixroos, line 32: dependency should also be avoided here

felixroos:
thanks so much for all the work! I think we should wait until this is stable before merging it. We should also make sure the tauri stuff does not break the webapp or adds additional dependencies. A pattern that I found useful to avoid dependencies is registering stuff from the outside, as used in `register` (pattern `registerSound` (webaudio) etc..

packages/midi/midi.mjs
vasilymilovidov: I'll see what I can do! it wast the first thing that worked for keeping notes more of less on the grid.

PR #636, opened 2023-07-03
More work on highlight IDs

mindofmatthew:
@felixroos, I've got this working! I split up the CodeMirror logic into two pieces of state: `mininotationLabels`, a RangeSet of all the mininotation symbols in the document that stores their ids, and then `visibleMininotationLabels`, a map of ids to their associated haps for a given query. The actual set of decorations is derived from these two pieces of state whenever either of them changes. This doesn't strictly change any of the behavior of your last version of the state field, but I think it makes each piece of the logic a little simpler to reason about. I can add comments as would be useful.

To test out my own understanding of the system, I went back and hooked up colored outlines, _and_ added a feature where the opacity of the outline indicates progress through the hap. I feel like this has been discussed before, but also happy to split this piece out into a separate feature branch if you all want to tweak implementation more.


felixroos:
amazing! Maybe the opacity stuff should indeed be a separate PR, as I'm not sure about the performance implications.. we'll have to do some testing to find that out. I guess the only problem left to solve is the failing tests, it seems the transpiler changes also changed some hap orderings.. I can look into it

mindofmatthew:
The opacity can definitely be a flag, extending the disable-highlights functionality. I'll play with that and do some performance tests in a separate PR.

Maybe related to tests, I noticed that the set of highlights returned from `getLeafLocations` seemed to sometime have duplicate locations. Does this address that too?

felixroos:
> The opacity can definitely be a flag, extending the disable-highlights functionality. I'll play with that and do some performance tests in a separate PR.

👍 

> Maybe related to tests, I noticed that the set of highlights returned from getLeafLocations seemed to sometime have duplicate locations. Does this address that too?

the tests were only failing because the order got messed up, now that https://github.com/tidalcycles/strudel/pull/637 is finally fixed, it now works! haven't seen doubled locations yet, I wonder how this could happen...

I've now refactored a bunch of things, so everything is nice and tidy. The old logic is gone now :)
I think that should be it, I'll merge it now and test again on the branch

Issue #506, opened 2023-03-03
stateless voicing function with anchors


felixroos:
I have long thought about this, but never opened an issue:
the current voicings function uses a global state, which can result in non deterministic behavior.
There could be an alternative function `voicing` (without s) that uses voicings without state.
Instead of using the voicing with the least distance to the last voicing (current behavior), the function could just use the voicing with the least distance to some anchor note, like

```js
"Dm7 G7".voiceBelow('c4').voicing('lefthand')
```

lefthand uses the following voicings:

```js
const lefthand = {
  m7: ['3m 5P 7m 9M', '7m 9M 10m 12P'],
  '7': ['3M 6M 7m 9M', '7m 9M 10M 13M'],
/*..*/
}
```

so the result would then be 

```js
"[c3,e3,f3,a3] [b2,e3,f3,a3]"
```

which is the same what the stateful version would do in this case.
the full set of functions could be:

- `voiceBelow` the top note of the voicing must be below the given note
- `voiceMax` the top note of the voicing must be below or equal to the given note
- `voiceAbove` the bottom note of the voicing must be above the given note
- `voiceMin` the bottom note of the voicing must be above or equal to the given note
- `voiceCenter` the average note must be as close as possible to given note

This also brings the advantage of being able to pattern those anchors, to let the voicings wander a bit, e.g.:

```js
"Dm7 G7".voiceBelow("<c4 f4>").voicing('lefthand')
```

edit: looking at https://github.com/tidalcycles/strudel/issues/504 this could maayybe even be implicit in the mini notation:

- `voiceBelow`: `"[Dm7 G7]:<c4 f4>".voicing('lefthand')"`
- `voiceAbove`: `"<c2 f2>:[Dm7 G7]".voicing('lefthand')"`

so if `voicing` receives a list, it applies `voiceBelow` if the second element is a note, and `voiceAbove` if the first is one. 
This would be an extreeemely cool feature to have when writing non boring chord progressions.

edit2: Another amazing capability of voiceBelow is that you can make a chord progression duck below a melody:

```js
"c a f e".superimpose(
  x=>"Dm7 G7 C^7@2".voiceBelow(x).voicings('lefthand'),
)
```

So in the above case, the chord progression will always stay below the melody!

felixroos:
poc https://codesandbox.io/s/stateless-voicings-g2tmz0?file=/src/index.js:260-376

felixroos:
[...]

PR #647, opened 2023-07-08
stateless voicings + tonleiter lib

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/506

implements a new algorithm for voicing chords, using the following functions:

## voicing

`voicing` is the only function that is not a control, but rather a function that consumes all the voicing related controls and renders them into actual notes. It accepts either strings or objects. Strings are interpreted as chords. Examples:

```js
voicing("C Am F G7"); // minimal example
voicing(chord("C Am F G7")); // alternative 1
chord("C Am F G7").voicing(); // alternative 2
```

## chord

`chord`s are just strings where the beginning is expected to be a valid pitch class (C, C#, Db, ... ) and the rest is considered to be the chord symbol (can also be empty, which is normally the case for major chords).

```js
chord("C Am F G7").voicing(); // uses chord symbols '', 'm', '' and '7'
```

## dictionary / dict

`dictionary` or short `dict` defines how chords are turned into individual notes, or in other words, how they are "voiced".
You can either pass a string or a dictionary:

```js
// 1. using predefined "triads" dictionary
chord("<Cm Fm>").dict("triads").voicing();

// 2. using custom dictionary
addVoicings("my-triads", {
  "": ["0 4 7", "4 7 12", "7 12 16"], // major chords (no symbol)
  m: ["0 3 7", "3 7 12", "7 12 15"], // minor chords via "m"
  // and so on..
});
chord("<Cm Fm>").dict("my-triads").voicing();

// 3. dictionary itself
chord("<Cm Fm>")
  .dict({
    "": ["0 4 7", "4 7 12", "7 12 16"], // major chords (no symbol)
    m: ["0 3 7", "3 7 12", "7 12 15"], // minor chords via "m"
    // and so on..
  })
  .voicing();
```

If `dict` is not set, the default dictionary `defaultDictionary` is used, which understands the most common chords (currently just triads and sevenths chords, but can be extended in the future).

In the dictionary, the numbers are semitones relative to the root note, so `['0 4 7', '4 7 12', '7 12 16']` in `C` would be `['C E G', 'E G C', 'G C E']`. These pitches have no octaves, so each voicing can be played in any octave.

## anchor

To decide which voicing to actually play for a certain chord, the `anchor` control is used.
It expects a single note to align the voicing to. It falls back to `E5`.
With the above dictionary and the default anchor, a C chord will always be 'G4 C5 E5', using the '7 12 16' option, because that voicing's top note is closest to `E5`.

The anchor is the actual "trick" to get voice leading without state, as a voicing does not need to know which voicing came last (which was the case with the old voicing algorithm).

## mode

The behavior of `anchor` is not fixed to always align the voicing to the top note. Instead, the `mode` control can be used to change that alignment to:

- below: picks voicing with top note <= anchor (default value)
- above: picks voicing with bottom note >= anchor
- duck: like below, but filters out top note if it matches anchor (good for harmonizing melodies)

## offset

The `offset` control applies an offset to the selected voicing. For example:

```js
chord("Cm")
  .dict({ m: ["0 3 7", "3 7 12", "7 12 15"] })
  .anchor("E5") // = default
  .voicing();
```

... the resulting voicing would be `G4 C5 E5`, which is the voicing with index 2 in the dictionary.
If we apply `.offset(1)`, the selected voicing index will be 2+1 = 3. Because there is no voicing with index 3, we wrap around to index 0, landing on `0 3 7`, giving `C5 E5 G5`. So with this particular set of voicings, the `offset` control acts controls the inversions, but that term would not apply for sets of voicings that do not include every rotation.
The above explanation sounds complicated, but when you use it, it makes much more sense..

## n

The `n` control selects an index out of the rendered voicing, which allows to play the voicing like a scale:

```js
n("0 1 2 3")
  .chord("Cm")
  .dict({ m: ["0 3 7", "3 7 12", "7 12 15"] })
  .anchor("G5") // <-- not default
  .voicing();
```

is equal to

```js
note("C5 E5 G5 C6");
```

Note that `n` = 3 wraps around to the first note of the voicing, but 1 octave higher.

edit: updated this whole post after the api changed during the development of this PR. Posts further down might still reference or iterate on the old version of this post.. The above functions are the actual up to date information


felixroos:
added another voicing feature: `voiceOffset`:

```js
voiceOffset("0 <[2 1] 2>")
  .chord("<Dm7 G7 C^7 A7b9>")
  .voiceMax("E4")
  .voicing('lefthand').piano()
```

`voiceOffset(0)` is the default, other values shift the voicing selection up (>0) or down (<0).
for n>voicings for the current chord, the voicing will be octavated.

felixroos:
and another feature: when `n` is set, `voicing` will use it to play the voicing as a scale. this is an alternative to `arp`, allowing to break the octave barrier + it's much more performant and logical imo. this also blurs the lines between scale and voicing, where voicing could also be used as a scale that sticks to an anchor. example:

```js
n(run(12))
  .chord("<Dm7 G7 C^7 A7b9>")
  .voicing('lefthand')
```

I am already having a blast using these new functions and I cannot wait to finish them.

felixroos:
I am still wondering if the voice dictionary (argument to voicing function) should also be a control..
With the current state, the dictionary can be patterned, but cannot be used as the structure, because `voicing` will use all the other controls, so it has to come last. Maybe there should be a `dictionary` control, then the voicing function would only expect the pattern as an argument, with the possibility of being used as `.voicing()`...

Also, another thought: related to https://github.com/tidalcycles/strudel/discussions/223 the voicing function could also accept an object with all the controls in it, like

```js
voicing({
  n: run(12),
  chord: "<Dm7 G7 C^7 A7b9>",
  voicing: "lefthand"
})
```

felixroos:
https://github.com/tidalcycles/strudel/assets/12023032/62cea237-b29f-4046-b02a-ad1596603775

felixroos:
`voicing` is now a function that takes only a pattern + `dictionary` is now just a control for setting the voicing dictionary.
I've also simplified to naming of the voicing controls:

```js
voicing(
    n(run(4)) // plays voicing like a scale
    .chord("<Dm7 G7 C^7 A7b9>") // which chord to play
    .dict("steps") // or dictionary
    .anchor("B5") // former called voiceMax / voiceBelow
    .add.squeeze(offset("0,3 1 2")) // how much to shift the voicing up or down
    .mode('max') // if set to below, the anchor note will be filtered out (for "melody ducking")
  ).piano()
```

alternatively, the above can also be written as:

```js
n(run(4)) // plays voicing like a scale
.chord("<Dm7 G7 C^7 A7b9>") // which chord to play
.dict("steps") // or dictionary
.anchor("B5") // former called voiceMax / voiceBelow
.add.squeeze(offset("0,3 1 2")) // how much to shift the voicing up or down
.mode('max') // if set to below, the anchor note will be filtered out (for "melody ducking")
.voicing().piano()
```

so the new control names are:

- dictionary / dict
- anchor
- (offset) already exists in the list, what is it?
- mode

i guess it's not important to namespace the controls for voicings, as they could be reused in another context (similar to `n` being reused for pitch, scale steps and now voicing steps)

felixroos:
this is now ready to merge. @yaxu ok with the naming of the new controls: https://github.com/tidalcycles/strudel/pull/647/files#diff-b1da86c00bc85a5a5848acac010d999cdfd66a9f97b9f107b09f70fa519c3b51R576 ?

> i guess it's not important to namespace the controls for voicings, as they could be reused in another context (similar to n being reused for sample number, scale step, pitch and now voicing steps)

yaxu:
Sorry for being slow!

Are these standard names in the world of harmony? I can't see them being used here, but I don't really know anything about voicing. https://en.wikipedia.org/wiki/Voicing_(music)

`offset` is used by superdirt to delay playback by the given number of seconds. `nudge` does the same, but is processed on the tidal side (by adding the nudge value to the osc timestamp). So we have redundancy there. I think this 'offset' is just copied from the original dirt implementation, and probably isn't documented anywhere and might not be used, so we could think about just removing it.

Are you keen on `offset`, or is there an alternative?

Overall no objections if you think these are the best names.

felixroos:
> Are these standard names in the world of harmony? I can't see them being used here, but I don't really know anything about voicing. https://en.wikipedia.org/wiki/Voicing_(music)

nope, only `voicing` and `chord` are terms in that world. the rest are just generic terms to steer the algorithm.
I am not aware of any terms that would accurately describe what they do, so i guess generics are ok.
If there are any musicologists out there reading this, knowing fitting terms, we can add them as aliases.
 
> `offset` is used by superdirt to delay playback by the given number of seconds. `nudge` does the same, but is processed on the tidal side (by adding the nudge value to the osc timestamp). So we have redundancy there. I think this 'offset' is just copied from the original dirt implementation, and probably isn't documented anywhere and might not be used, so we could think about just removing it.
> 
> Are you keen on `offset`, or is there an alternative?

Not specifically keen on it. But even if there was a naming collision, it would not be a problem, as `voicing` will "consume" its controls and only output notes (+ everything that is not related to voicings). Example:

```js
n("0 1 2 3").chord("Cm").voicing().s("gm_epiano1")

// alternative syntax:
voicing(
  n("0 1 2 3").chord("Cm")
).s("gm_epiano1")
```

In the above example, `n` is used as voicing steps. After `.voicing()`, it is gone, so `gm_epiano1` won't receive different values for `n`. The same would apply for all the other voicing related controls, including `offset`. So they could be seen as being only in a local namespace. This syntax could be generally useful for more complicated functions like that.

felixroos:
FYI I've now added a more in depth description of all the new functions at the [top of this PR](https://github.com/tidalcycles/strudel/pull/647#issue-1795093726)

yaxu:
Ah nice! Sorry I hadn't looked into the details it, the scoping is nice !

PR #653, opened 2023-07-22
ireal voicings

felixroos:
adds ~611~ 638 voicings for all chord symbols supported by ireal pro!
all the voicings can be explored here: https://f5lv24.csb.app/

more info on where these voicings come from:

the source code for the scraper is here: https://codesandbox.io/s/ireal-midi-scraper-forked-gjz2mr
the scraper takes a midi file as generated by ireal pro (you can export any playalong as midi).
To scrape all chords, I created a custom tune that contains all chords and goes through all keys, which should contain all / most voicings of the software. 
In the sandbox, the files "Allchords_mediumswing_noext.mid" "Allchords_mediumswing_ext.mid" are the midi files exported by ireal, one without and one with chord extensions.
The scraper parses the midi via `@tonejs/midi` and knows the order of chords.
Then it collects all note offs as groups and calculates the voicing for each chord.
All unique voicings are then collected in the format strudel expects. 
The file `ireal.mjs` contains 2 different sets of voicings, one without and one with added extensions.

They can be used with `.dict('ireal')` or `.dict('ireal-ext')`

Btw chord voicings are not copyrightable...


felixroos:
todo: sort by top note

PR #656, opened 2023-07-29
[Bug Fix] Account for numeral notation when converting to midi

daslyfe:
Discovered this bug when trying to use numeral notation with midi for example: 
`"12 7".note().midi('mymididevice').midichan(1)` 
would throw an error like 
`[cyclist] error: not a note: "7"`


felixroos:
great thanks!

PR #657, opened 2023-07-29
[Bug Fix] Midi: Don't treat note 0 as false

daslyfe:
This bug caused note 0 to be ignored.

I suspect there might be similar issues in other places in the app.
 For instance: if you do 
`"0".note().sound("triangle").pianoroll()`
it will display nothing in the piano roll and play a c2 and 
`"36".note().sound("triangle").pianoroll()`
will display in the piano roll and play a c2 as well


felixroos:
thank you for that fix!

PR #661, opened 2023-08-10
basic reference with one page per function

felixroos:


felixroos:
hmm the prebuilt reference pages are 10MB in total, just because there are so many: a total of 229 paths, each having ~40kB.. I wonder if that 10MB per build will exceed some quota or slow down the build. if yes, it should make more sense to find a way to do this dynamically.

felixroos:
maybe it makes more sense to have a single reference page with anchor links? would be significantly smaller + no extra sidebar would be needed (can use right sidebar) + full text search would be much more useful

hellocatfood:
I think the anchor links is one solution, assuming that it works better than the current solution.

However, I still think having one function per page would be better, as each page could have more information and more examples on, including links to similar/related functions. See how pure data handles that below

image


With regards to build size, could the docs be built separately?

felixroos:
> However, I still think having one function per page would be better, as each page could have more information and more examples on, including links to similar/related functions. See how pure data handles that below

Yep it's certainly better..

> With regards to build size, could the docs be built separately?

It is already built separately but you'd still have to download everything as part of the desktop app

It would be possible to have a URL like `strudel.tidalcycles.org/ref?fast` which would result only in a single html file.. 
I'll try that

Issue #119, opened 2022-05-28
[osc] mysterious clock drifting


felixroos:
Under some unknown circumstances, the osc messages will arrive late in supercollider. 
On lowend machines, this seems to happen more easily, though it could be a separate problem.
There seems to be some clock drifting going on, most likely that osc uses the system clock and strudel uses Date.now (on page load) +  audio time since since page load. So far, I was only able to get late messages after a longer time of inactivity without refreshing the repl. after a refresh, the clocks are aligned again.

The clock drifting could be avoided if .osc would use performance.now instead of audio time.

bwagner:
This happened on a MacBook Pro with chip Apple M1 Max, memory 32 GB running Ventura 13.2.1. Drifting started after around 45 seconds.

felixroos:
same drifting seems to happen with midi too. the midi clock fell ~15ms behind after roughly a minute. Midi uses WebMidi.time which is based on performance.now() and the scheduler uses AudioContext.currentTime, so these clocks are probably also not running in sync

felixroos:
the scheduler is clock agnostic, so it might make sense to test what happens when it uses performance.now(). My guess is that midi will run perfectly but web audio will get out of sync..

felixroos:
this could be the solution for the clock drift: https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/getOutputTimestamp

felixroos:
fixed with https://github.com/tidalcycles/strudel/pull/666

felixroos:
this is not fixed.. it seems to be related to accumulating float rounding errors. this seems to only happen for certain cps numbers: .5 1 2 is fine, while .75 is not.. The problematic bit is most likely:

```js
const begin = this.lastEnd;
this.lastBegin = begin;
const end = round(begin + duration * this.cps);
this.lastEnd = end;
const haps = this.pattern.queryArc(begin, end);
```

the calculation `duration*this.cps` does not have perfect precision, so each tick the error is accumulating, creating a larger and larger drift as time passes. this should be fixable with fractions

yaxu:
It doesn't necessarily need fractions, you can avoid accumulating rounding errors with a calculation like: 
```javascript
begin = lastend
end = time_of_cps_change + (number of frames_since_change * time_per_frame)
```

jarmitage:
workaround
```js
const loopAx = register('loopAx', 
[...]

PR #666, opened 2023-08-17
fix osc bundle timestamp glitches caused by drifting clock

daslyfe:
The clock from the audio context can drift over time, for me it is about +1 seconds every 20 minutes. 

When using the startedAt time as a reference, the drift can cause the ts value to be inaccurate and cause timing issues when sending the osc bundle. 

Instead, calculate the offset from the time values provided by the trigger and add them to the current time to ensure that the timestamp sent to the bundle is accurate.


felixroos:
thanks! this fixes https://github.com/tidalcycles/strudel/issues/119 which is the oldest reported bug

PR #668, opened 2023-08-19
Midi time fixes

daslyfe:
- implement the note offset more accurately using the method[ recommended by the api](https://webmidijs.org/api/classes/Output#playNote) "If time is a string prefixed with "+" and followed by a number, the message will be delayed by that many milliseconds." 

- use the Note constructor as recommended by the [api docs](https://webmidijs.org/api/classes/Note)

- fix small issue where listeners would sometimes not trigger or trigger many times and spam the console with the device initialization log message

- default to first IAC device if no devices is specified in pat.midi() and an IAC device exists. 

- adjust the duration offset slightly (from -5 to -10) to improve note glitching issues due to inconsistent duration/note off messages from the WebMidi API (this seems to be  worse in firefox than chromium

### DIFF
--- a/packages/midi/midi.mjs
+++ b/packages/midi/midi.mjs
@@ -6,8 +6,8 @@ This program is free software: you can redistribute it and/or modify it under th
 
 import * as _WebMidi from 'webmidi';
 import { Pattern, isPattern, logger } from '@strudel.cycles/core';
-import { getAudioContext } from '@strudel.cycles/webaudio';
 import { noteToMidi } from '@strudel.cycles/core';
+import { Note } from 'webmidi';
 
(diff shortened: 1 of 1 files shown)


felixroos:
amazing! have you tested this new timing logic over longer periods of time? i don't fully remember why the old logic became what it is, but there were some long term timing problems (clock drifting?)

daslyfe:
> amazing! have you tested this new timing logic over longer periods of time? i don't fully remember why the old logic became what it is, but there were some long term timing problems (clock drifting?)

It shouldn't drift out of time, because it just sends an offset value at the point that the note was triggered (instead of a full timestamp). I have been working on a music sketch all day that uses webmidi, osc, and the audio engine and they all seem to be keeping consistent time with each other.

Issue #191, opened 2022-08-14
support waveformN format


felixroos:
it would be nice to have waveforms like `sine8` `sawtooth16` etc. of the format `waveformN` where waveform is one of `sine`, `triangle`, `sawtooth`, `square` and `N` a number between 1 and 128, specifying the number of harmonics used. It is a quick and handy way to get a filtered waveform without actually needing a filter, because it's additive synthesis. This can probably with a [PeriodicWave](https://developer.mozilla.org/en-US/docs/Web/API/OscillatorNode/setPeriodicWave)

felixroos:
here's a codesandbox that implements the basic idea: https://codesandbox.io/s/waveformn-createperiodicwave-www4m9?file=/src/index.mjs

PR #674, opened 2023-08-24
control osc partial count with n

felixroos:
for any of the simple waveforms, sawtooth, triangle, square (sine), this allows controlling the number of partials with the `n` control:

```js
note("c a f e")
.s("sawtooth:8") // <--- sawtooth wave with 8 partials
```

This allows creating less harsh waveforms with additive synthesis directly, without needing a filter, which is more to write.

This is actually a breaking change, because `n` could prior be used to control the oscillator pitch, as a synonym for `note`.
I think this synonym, which only works with oscillators is more confusing than helpful, as it sometimes works ( waveforms ) and sometimes not ( samples: picks another sample ). The behavior of `n` to choose partial count for oscillators complements the behavior of `n` to pick a sample variant, as it is also timbre related..

@yaxu what do you think? I think tidal also has this separation of `n` for samples and synths? 
If you don't think this is a good idea, I could also add a `partials` control instead...

edit: the problem with an extra control for this is that you cannot use it with ":" notation, so you'd have to write `s("sawtooth").partials(8)` instead of `s("sawtooth:8")` because the second item in `s` is always bound to `n`.
Right now, that second item is not really usefull for waveforms

also, it might make sense to limit the number of partials to 64 or something

fixes https://github.com/tidalcycles/strudel/issues/191
related: https://github.com/tidalcycles/strudel/discussions/395

### DIFF
--- a/packages/superdough/helpers.mjs
+++ b/packages/superdough/helpers.mjs
@@ -6,17 +6,6 @@ export function gainNode(value) {
   return node;
 }
 
-export const getOscillator = ({ s, freq, t }) => {
-  // make oscillator
-  const o = getAudioContext().createOscillator();
-  o.type = s || 'triangle';
-  o.frequency.value = Number(freq);
-  o.start(t);
-  //o.stop(t + duration + release);
-  const stop = (time) => o.stop(time);
-  return { node: o, stop };
-};
-
 // alternative to getADSR returning the gain node and a stop handle to trigger the release anytime in the future
 export const getEnvelope = (attack, decay, sustain, release, velocity, begin) => {
   const gainNode = getAudioContext().createGain();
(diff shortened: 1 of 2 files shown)


yaxu:
So I guess the idea is to let the synth to decide what `n` means in terms of the main way it varies the timbre? I guess some might still map it to `note` in that case, unless that's just to confusing..

felixroos:
> So I guess the idea is to let the synth to decide what `n` means in terms of the main way it varies the timbre? I guess some might still map it to `note` in that case, unless that's just too confusing..

if synth means the simple waveforms, not really.. instead decide that simple waveforms will always use n for the number of partials. if n is zero, then "infinite" partials will be used (no additive synthesis). but the fact that waveforms / synths and samples are built on top of "sounds" as in https://strudel.tidalcycles.org/technical-manual/sounds, it means that a sound can decide what happens with any of the controls passed to it.. but in the majority of cases, people won't use that sound api. but in the future, we could use the sound api to build different synths that maybe use n in a different, timbre controlling way, yes

felixroos:
ok to merge this then?

yaxu:
Sorry missed the question but the note / n confusion has been a pain point for many and this seems like an improvement..

yaxu:
sure keep it moving :)

PR #677, opened 2023-08-25
Scope

felixroos:
adds oscilloscope for time and frequency domains

image

image

todo:

- [x] move to webaudio or superdough package, as it depends on superdough
- [x] move fft control to scope params => no


felixroos:
I guess this is now complete.. i have more ideas to improve on this but i think it's already cool as is. follow up issue: https://github.com/tidalcycles/strudel/issues/678

PR #683, opened 2023-08-29
Wave Selection and Global Envelope on the FM Synth Modulator

Bubobubobubobubo:
This commit is adding a gain envelope (ADSR) on the FM modulator. It also allows the user to choose the modulator waveform from the basic WebAudio `OscillatorNode` types. It exposes the following parameters:

- `fmwave: fmWaveform = 'sine'` (all base oscillators are supported)
- `fmattack: fmAttack = 0.001`
- `fmdecay: fmDecay = 0.2`
- `fmsustain: fmSustain = 0.001`
- `fmrelease: fmRelease = 0.1`
- `fmvelocity: fmVelocity = 1`
- `fmenv: fmEnvelopeType = 'linear'`

There are two different envelope types: `linear` and `exp`. Exponential is still a bit clunky but it generates very nice timbres if you push the decay just a bit more than usual.


felixroos:
cool! the
I've changed the default envelope to be no envelope, so longer sounds wouldn't get decayed after .2 seconds.
This will also improve performance, as no envelope needs to be calculated if you don't set one..

I've also added the missing controls for strudel + added some missing params to the exp envelope, which sounds pretty nice btw

i think this now looks good!
ok?

Bubobubobubobubo:
OK!

PR #685, opened 2023-08-30
Create Midi Integration for Tauri Desktop app

daslyfe:
The timing is more stable/less jittery than the webmidi implementation. Thanks @vasilymilovidov for providing a POC to reference.


felixroos:
great stuff, the timing seems rock solid!
i see a problem with using desktopbridge in the midi package. The midi package shouldn't depend on tauri / desktop stuff, as it has to work independently. 
We can safely add `@tauri-apps/api` as a dependency to the website instead.
Then we can decide in `Repl.jsx` to either import `@strudel.cycles/midi` or the desktopbridge based on the outcome of `isTauri()`. Both `@strudel.cycles/midi` and the desktopbridge can then just define `Pattern.prototype.midi` and only the relevant one will be used.

edit: I've now tried that approach and it seems to work

felixroos:
one more little nice to have idea: The tauri app already logs the available midi devices to the terminal, maybe there is a way to also log those to the repl console, similar to how it happens with web midi, so you'd see the correct names also when using the desktop app regularly without terminal

daslyfe:
> edit: I've now tried that approach and it seems to work

Cool, I agree with this approach I think it makes sense.

>one more little nice to have idea: The tauri app already logs the available midi devices to >the terminal, maybe there is a way to also log those to the repl console, similar to how it > happens with web midi, so you'd see the correct names also when using the desktop > > app regularly without terminal

I think this is a good idea, I was planning to see if this can be done (I think it should be, I know that Tauri can communicate in both directions) and make a followup PR.

PR #692, opened 2023-09-04
Adding filter envelopes and filter order selection

Bubobubobubobubo:
This commit is proposing the addition of linear or exponential filter envelopes:
- `lpf(5000).to(500).over(0.2)`

It also makes it possible to select the filter order (12db/Oct or 24db/Oct): 
- `lpf(5000).order("<12 24>")`

The definitive API and parameter names are not fixed but this is working and it sounds great 😄


felixroos:
great, filter envelopes were really missing so far. Maybe go the full way and just add adsr to the filter? The naming system could be similar to how it works with `f` and `q` suffixes:

- `lpattack` / `lpa` , `lpdecay` / `lpd` , `lps` / `lpsustain` , `lprelease` / `lpr`
- `bpattack` / `bpa` , `bpdecay` / `bpd` , `bps` / `bpsustain` , `bprelease` / `bpr`
- `hpattack` / `hpa` , `hpdecay` / `hpd` , `hps` / `hpsustain` , `hprelease` / `hpr`

there could also be `lpadsr` / `bpadsr` / `hpadsr` ? also not sure if release is really needed in the filter

Order being a number kind of suggests it supports arbitrary numbers, which might be confusing, not sure what a good alternative is. Maybe default to 24dB entirely? or add separate controls like `lp2` (because it's a 2nd order low pass filter)

Bubobubobubobubo:
This is now working with the described API. Order is also taking `12db` or `24db` only as argument, defaulting to `12db`. There is an additional argument to specify the depth of modulation.

felixroos:
this is starting to look good

still unsure about the naming:

- should `fenv` be controllable separately for each filter? then there would be `lpenv` `hpenv` `bpenv`
- rename `fenv` to something more expressive? it's also easy to confuse with `fmenv`
- order sounds too generic + it's weird that there is only one value that changes something and you have to always type '24db'. maybe replace order with ftype (or lptype / hptype / bptype) where other filter types might be added in the future?
- other filter env controls need doc comments too to appear in the reference & autocomplete. might be a bit repretitive to also add them to the fx docs page

Bubobubobubobubo:
Your suggestions are now implemented. Examples are covering all new keywords (mostly duplicate of lowpass examples changing filter type). Tests have been re-generated, etc..

PR #693, opened 2023-09-04
Adding vibrato to base oscillators

Bubobubobubobubo:
This commit is adding vibrato control to the base synth:
- `vibrato` to set the vibrato speed (in _hz_).
- `vdepth` to set the vibrato depth (**uncapped!**)

~~It also adds pitch slide:~~
~~- `slide` (if non `0`): returns `slide` * `freq`~~
~~- `slide_speed`: `sustain` / `slide_speed`~~

It's now very easy to give some life to these however very static oscillators.


felixroos:
nice! some feedback:

- I think the default value and scaling of vdepth is too high. something like `s("triangle").vibrato(2)` creates a really harsh pitch sweep. the default value is 100 and the scaling is *100, so the default changes pitch by 10000Hz .. I'd maybe scale it to semitones + set the default to 1 or something like that
- maybe find a better name than vdepth? maybe shorten it to `vib` and `vibmod`. `vib` could also accept ":" notation like `.vib("1:2")` setting both values.
- probably makes sense to keep `getOscillator` thin and implement it directly in synth.mjs (similar to fm, which is also a vibrato).
- i don't really get how slide is supposed to be used? I always get a pitch jump instead of a slide. also not sure why / how slide is coupled with vibrato and sustain
- I'd rename slide_speed to slidespeed, as controls never use _



but that

felixroos:
maybe also try to match zzfx's slide property..

Bubobubobubobubo:
Changes:
- `vib` and `vibmod` are now the parameter names. 
- The modulation is less extreme by default. `vibmod` is introducing gentle modulation between `0.0` and `1.0` and can go way beyond for more drastic vibrato that borders on FX.
- `vib` is still in hertz

Slide has been renamed `pitchJump`, same terminology used by ZzFX.

PR #694, opened 2023-09-04
Direct OSC Support in Tauri

daslyfe:
Negates the need to run a separate websocket server for communicating with supercollider when using the desktop app.


felixroos:
amazing, just tested it and works like a charm! 

only thing I noticed is that when you exit sclang while the desktop app is running, this error is logged to the console:

```sh
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 61, kind: ConnectionRefused, message: "Connection refused" }', src/oscbridge.rs:67:37
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

after that, it only works again after the app has been restarted. maybe there could be some reconnect polling logic when the connection is lost / not found. but I guess that could also be a separate PR.

daslyfe:
It is not a good idea to make the thread crash if a message fails to send. I fixed the behavior. Does that fix things on your end as well?

felixroos:
> Hmm it looks like It is forcing the thread to panic if a message fails to send (which we probably dont want). I fixed the behavior. Does that fix things on your end as well?

yep works! getting "OSC Message failed to send, the server might no longer be available" and it also works again when I restart it! so merge this then?

daslyfe:
yup! go for it

PR #695, opened 2023-09-04
Rust WebAudio for Tauri Desktop App

vasilymilovidov:
Figuring out rust webaudio api for Tauri

### DIFF
--- a/packages/desktopbridge/index.mjs
+++ b/packages/desktopbridge/index.mjs
@@ -6,3 +6,4 @@ This program is free software: you can redistribute it and/or modify it under th
 
 export * from './midibridge.mjs';
 export * from './utils.mjs';
+export * from './webaudiobridge.mjs'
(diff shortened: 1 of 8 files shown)


vasilymilovidov:
Moving along with webaudio for tauri. 

What roughly works:
- playing waveforms
- playing samples from computer
- sample speeds 
- filters
- beginning, end, and functions that utilize them
- loop, loopBegin, loopEnd

vasilymilovidov:
> Interesting stuff! I wonder if it would make sense to build the rust part as wasm and use that on the web again :D it might actually be faster as it will only use a single AudioWorklet, not sure though

as I understand it, this webaudio api rust library doesn't support wasm. so it would lead us to a crossroads of sound engines once again haha.

vasilymilovidov:
after finishing this, I want to take another look at glicol to see if it's possible to use in strudel without its livecoding syntax

PR #699, opened 2023-09-14
Midi in

felixroos:
allows using midi cc to control a mutable variable that itself is a pattern!

example: 

```js
let apc = await midin('APC MINI')
  
n(run(apc(54).range(1,12).round()))
  .chord("<Cm D>/8")
  .jux(rev)
  .offset(apc(55).range(-10,4).round())
  .voicing()
  .add(note(apc(48).range(0,12)))
  .lpf(apc(49).range(200,2000))
  .clip(apc(50).range(0.01,4))
  .room(apc(51).range(0,2))
  .delay(apc(52).range(0,1))
  .sometimesBy(apc(53), add(note(12)))
  .piano()
```

maybe note input would also be a good fit here

demo: https://www.youtube.com/watch?v=e2-Sv_jjDQk

another, minimal example:

```js
let cc = await midin('APC MINI')
s("hh*8").speed(cc(48).range(.5,1))
```

this will allow you midi device to control the hihat speed between .5 and 1


maximeborges:
Is this still supposed to work I'm getting a bunch of JS errors in the console with the following code based on your minimal example (for an APC Mini MK2, but also for a custom controller):
```
let cc = await midin('APC mini mk2')
s("hh*8").speed(cc(48).range(.5,1))
``` 
```
[query]: can't access property 48 of undefined 2 [index.CVox9G2w.js:7:5653](https://strudel.cc/_astro/index.CVox9G2w.js)
[query]: can't access property 48, y[c] is undefined
```

PR #706, opened 2023-09-20
Adding vibrato to Superdough sampler

Bubobubobubobubo:
Modulating the `playbackRate` to simulate a vibrato on the sampler.


felixroos:
the scaling of vibmod is different compared to how it is scaled on the oscillator.. Both should match the same standard I'd say

felixroos:
> the scaling of vibmod is different compared to how it is scaled on the oscillator.. Both should match the same standard I'd say

now fixed

Issue #709, opened 2023-09-26
Support MIDI Clock


atfornes:
Sending MIDI clock messages would be really useful to synch with MIDI instruments.

WebMIDI clock offers [sendClock function](https://webmidijs.org/api/classes/Output/#sendClock), that should be called 24 times per quarter note.

atfornes:
Midi uses 248 as number for clock signal. Unfortunately, sending this message with current strudel midi functionality does not work:
`note("248*24").midi()`

atfornes:
Already discussed in #192

PR #710, opened 2023-09-27
add midi clock support

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/709

usage:

```js
stack(
  s("hh*4"),
  midicmd("clock*48").midi('IAC-Treiber Bus 1')
)
```

This allows manipulating the clock itself with a pattern :)
Additionally, it will fire a midi start event when the repl is started to make sure the clock is in phase.
On the very first start, it might behave weird when midi is not initialized yet...

edit: adjusted for new syntax

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -1066,6 +1066,7 @@ const generic_params = [
    */
   ['waveloss'],
   // TODO: midi effects?
+  ['clock'],
   ['dur'],
   // ['modwheel'],
   ['expression'],
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -1066,7 +1066,7 @@ const generic_params = [
    */
   ['waveloss'],
   // TODO: midi effects?
-  ['clock'],
+  ['midicmd'],
   ['dur'],
   // ['modwheel'],
   ['expression'],
--- a/packages/midi/midi.mjs
+++ b/packages/midi/midi.mjs
@@ -78,6 +78,20 @@ function getDevice(output, outputs) {
   return IACOutput ?? outputs[0];
 }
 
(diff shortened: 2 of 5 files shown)


yaxu:
Nice! I have limited experience with midi clock, but did find it can be nice to send midi start more often, say every cycle. Some synths don't like this though. It can also be fun to pattern midi starts.

Superdirt uses a 'midicmd' control to specify the different midi commands. https://tidalcycles.org/docs/configuration/MIDIOSC/midi/#synchronising-midi-clock-via-tidal

This mirrors the supercollider midi api https://doc.sccode.org/Tutorials/A-Practical-Guide/PG_08_Event_Types_and_Parameters.html#MIDI%20output

felixroos:
> Nice! I have limited experience with midi clock, but did find it can be nice to send midi start more often, say every cycle. Some synths don't like this though. It can also be fun to pattern midi starts.

i should've looked at the tidal doc first :) yep that midicmd control sure looks nice! I've changed it to now allow:

```js
stack(
  note("<c eb f g>*2").ply(2).s("gm_acoustic_bass").sub(note(12)),
  midicmd("clock*24*2,<start stop>/2").midi('IAC-Treiber Bus 1')
)
```

I've added a an alias `clock` for `midiclock`, because the namespace seems redundant, as the control is already namespaced...

felixroos:
I've now adding automated start stop messages on top of it, because at least ableton does not like when the midi clock ends without sending a stop. the start could theoretically be left to the user, but it seems weird to only send the stop.

felixroos:
ok starting automatically is a good idea, because my keystep will treat a second start message as toggle, so a continuous playback wouldn't be possible without a one time start command

PR #714, opened 2023-09-30
widgets

felixroos:
image

- adds `slider(value, min?, max?)` that will automatically spawn a slider in the code. 
- the slider is inlined with the code so it will move with edits. 
- when the slider is moved, the value in the code will change as well, so the code will always be in sync
- the slider function returns a pattern!

this might still contain a tiny bit of jank but maybe it's already good enough for testing as a first version.

short video demo:

https://github.com/tidalcycles/strudel/assets/12023032/f06b227e-ff8a-4815-b616-61dfe8476301


felixroos:
I'll merge this now as it seems bugfree.. the will certainly be a follow up to improve styling / add different options etc..

PR #716, opened 2023-10-01
support mininotation '..' range operator, fixes #715

### LINKED ISSUE #715: Mininotation ranges
Tidal has [0 .. 3] as shorthand for 0 1 2 3. [0, 0.5 .. 2] might also be nice for 0 0.5 1 1.5 2 but tricky when , is already taken.
In tidal it has to be in its own subsequence, "0 .. 3" and "[0 .. 3] 4" are ok, "0 .. 3 4" and "[0 .. 3 4]" are syntax errors. <0 .. 4> works but (surprisingly) is the same as [0 .. 4]. These edge cases are probably bugs to fix though.

In tidal it works with notes too, e.g. `"a .. e"`, and going backwards `"10 .. 2"`

Unfortunately javascript doesn't have a built-in range operator or type inference which makes some of this a bit hard to implement.

Contributes to #30

yaxu:


felixroos:
great stuff! looks good, I only see 2 console.logs still around

yaxu:
@felixroos ah thanks for catching that!

PR #717, opened 2023-10-01
Add support for using samples as impulse response buffers for the reverb

vasilymilovidov:
This PR adds a function `ir()` that accepts the names of samples to load as an impulse response for the reverb. the `size()` function creates a new buffer of the requested size. 
```
s("bd").room(0.5).size(2).ir("shaker_large")
```
This also works:
```
s("bd").room(0.5).size(2).ir("shaker_large:2")
```

If the `ir()` function is not applied, the regular impulse response generation is used.


felixroos:
after the first reverb PR is now merged, it seems it has created a bunch of conflicts here.. should I try to resolve them or do you want? :P

vasilymilovidov:
> after the first reverb PR is now merged, it seems it has created a bunch of conflicts here.. should I try to resolve them or do you want? :P

if I'm not mistaken, right now this is just formatting in reverbGen file, and the rest is just my additions

felixroos:
I had to do another merge, where I also removed the setIR function. it still seems to work as intended or did I remove something I shouldn't have?

vasilymilovidov:
> I had to do another merge, where I also removed the setIR function. it still seems to work as intended or did I remove something I shouldn't have?

setIR was indeed redundant I guess, but there's a problem now with getting a url out of an object. I get this when I try to use samples like 'ocarina' or 'harmonica'  
```
[cyclist] error: url.replace is not a function (4)
```

felixroos:
works like this, not sure if this is as it was before..

vasilymilovidov:
yeah, everything seems to work great!

felixroos:
so ready for merge?

vasilymilovidov:
I think so, yeah

PR #719, opened 2023-10-02
Ableton link integration

daslyfe:
Currently working: 
-Stop/Start Sync from both directions
-Changing CPS of the Abelton link clock, although I was unable to figure out how to get the CPS of the "main" clock in Strudel. Comments in code.

needs to be implemented: 
- hooking up phase syncing/ CPS changes in strudel. I left comments in the code where I thought it could be added, I just wasn't sure how to do it. could use some advice here or feel free to just take over the pr.

### DIFF
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -18,24 +18,30 @@ export class Cyclist {
     this.lastEnd = 0; // query end of last tick
     this.getTime = getTime; // get absolute time
     this.onToggle = onToggle;
-    // this.abeLinkListener = listen('abelink-event', async (e) => {
-    //   const payload = e?.payload;
-    //   if (payload == null) {
-    //     return;
-    //   }
-    //   const { play, bpm, timestamp } = payload;
-    //   console.log('play');
+    this.start_timer;
+    this.abeLinkListener = listen('abelink-event', async (e) => {
+      const payload = e?.payload;
+      if (payload == null) {
+        return;
+      }
+      const { play, bpm, timestamp } = payload;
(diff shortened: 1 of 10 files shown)


felixroos:
this is great, tested it with ableton and it works! sometimes the phase is not picked up correctly but a restart fixes it. I am also sometimes not able to start strudel, it just does nothing when i evaluate, but it works with some mashing.

> Changing CPS of the Abelton link clock, although I was unable to figure out how to get the CPS of the "main" clock in Strudel. Comments in code.

now setting the cps accordingly.. the clock desync problem is appearing here as well when setting the cps to not so round floats, like .75. .5, 1 and .8 seem to work. this is the root of the problem: https://github.com/tidalcycles/strudel/issues/119#issuecomment-1743644467 this should probably be fixed in a separate PR.

> hooking up phase syncing/ CPS changes in strudel.

I've added a setPhase to the clock, but it seems the `phase` of the payload and the phase of the clock do not really fit together. `clock.getPhase()` returns the seconds since the playback started, while `phase` from the payload is always a very small number, often 0. @daslyfe do you know what that phase is relative to?

felixroos:
hm ok it seems the phase is also desyncing when strudel is open and paused for a bit. only able to sync it again when restarting strudel

daslyfe:
" @daslyfe do you know what that phase is relative to?"

The phase for link represents where the clock is in a single bar at the set tempo (it should count from 0 -4), so I think this can be translated in strudel to be the length of a single cycle. So a phase of 0 should be the start of the cycle, and a phase of 2 should represent the middle of a cycle.

PR #730, opened 2023-10-08
Reverb: autodetect sample rate

Bubobubobubobubo:


felixroos:
now fixed in https://github.com/tidalcycles/strudel/pull/732 (could not edit because it was on your main branch, so I couldn't check it out..)

Issue #406, opened 2023-02-03
add shabda shortcut


felixroos:
instead of `await samples('https://shabda.ndre.gr/bass:4,hihat:4,rimshot:2.json?strudel=1')`:

```js
await samples('shabda:bass:4,hihat:4,rimshot:2')
```

or maybe even

```js
await shabda('bass:4, 'hihat:4', 'rimshot:2')
```

also find a format for voice

ilesinge:
Here are some ideas for speech (with no params shabda defaults to en-GB female voice):
```javascript
await samples('shabda/speak:strudel,rocks')
```
and with language param:
```javascript
await samples('shabda/speak/fr-FR:strudel,pour_tous')
```
finally with gender:
```javascript
await samples('shabda/speak/fr-FR/m:strudel,pour_tous')
```
It's a bit more difficult for the dedicated function. In order to differenciate between  words and params I would rather do:
```javascript
await shabda('chocolate:3,pudding:2')
```
and (with optional language and gender params)
```javascript
await shabdaSpeak('strudel,pour_tous', 'fr-FR', 'm')
```
What do you think? Let me know if you want me to try to implement some of them.

felixroos:
> What do you think? Let me know if you want me to try to implement some of them.

I like the approach with `samples` more because it's less mystified. it's also analogous to the `github:` shortcut. `/speak` for voice looks good. or maybe name it `speech` as shabda does in its url? PR welcome!

ilesinge:
First PR on strudel :)
Let me know if it fits your idea and code style.

PR #740, opened 2023-10-14
Add shabda shortcut

ilesinge:
Implements https://github.com/tidalcycles/strudel/issues/406


felixroos:
thanks! To fix the test, I think you need to run `pnpm snapshot` to regenerate the test output for the example you've added

ilesinge:
Just did, thanks for the hint :)

ilesinge:
Thanks!

PR #741, opened 2023-10-18
[Bug Fix] chooseWith: prevent pattern from stopping audio when selection is >= 1 or < 0

daslyfe:
Also documented the function


felixroos:
thanks :)

Issue #569, opened 2023-05-13
Links to manual bring up the REPL?


yaxu:
For example if I follow a link to:
https://strudel.tidalcycles.org/technical-manual/alignment/

The last strudel pattern I played at comes up instead.

If I do a hard page refresh then the manual appears.

felixroos:
hm weird stuff happening.. maybe some cache interference going on. Maybe just link https://github.com/tidalcycles/strudel/blob/a03541295627ee7f1a6df545a61dd9ab7b68d421/website/src/pages/technical-manual/alignment.mdx for now :-/

felixroos:
it works without the slash at the end: https://strudel.tidalcycles.org/technical-manual/alignment

felixroos:
found the problem, described here: https://github.com/vite-pwa/astro/issues/23#issuecomment-1773447037

felixroos:
fixed with https://github.com/tidalcycles/strudel/pull/753

PR #743, opened 2023-10-20
fix: trailing slash confusion

felixroos:
this hopefully fixes https://github.com/tidalcycles/strudel/issues/569


felixroos:
possibly related: https://github.com/vite-pwa/astro/issues/23

Issue #750, opened 2023-10-21
idea: support international alphabets in sample names


ilesinge:
As a non-English natively speaking user, I would like to have sample names in my mother tongue (especially spoken words).

Currently, only alpha-numeric characters are allowed in the [parser](https://github.com/tidalcycles/strudel/blob/main/packages/mini/krill.pegjs#L106).

This would be particularly interesting in combination with the speak function and with shabda's speech mode, which both do direct step name to voice generation and thus cannot be used currently with some non-English languages in mininotation.

Looking at [this pegjs issue](https://github.com/pegjs/pegjs/issues/226#issuecomment-32122017) and after experimenting that the [Unicode Character Categories such as \p{Letter}](https://dev.to/tillsanders/let-s-stop-using-a-za-z-4a0m) are not supported in pegjs, it would seem that the most obvious, if not elegant, solution, would be to add all alphabets in the parser such as in [the example JavaScript parser](https://github.com/pegjs/pegjs/blob/93d3c6a2509f947f57d5728c15d8b96ceedef28f/examples/javascript.pegjs#L387).

felixroos:
yep i'd also love to use german special chars sometimes + using emojis would be cool too

ilesinge:
The pull request adds german special chars (but excludes emojis), if you want to have a look (not super satisfied about it but I go around pegjs limitations)

felixroos:
fixed in https://github.com/tidalcycles/strudel/pull/751

PR #751, opened 2023-10-21
mini notation: international alphabets support

ilesinge:
Implements #750 

Note that without reducing the max error message's length to 180 in the Repl (while keeping the end of the error message), the screen becomes unusable when one use a character that is not included (such as "$"). The full error message can still be found in the console as seen below.
It could be possible to do the same length reduction with the error in the console.

image


felixroos:
looks like there are ways to customize the peggy error message: https://peggyjs.org/documentation.html#error-messages  maybe it is customizable enough to allow not printing all these characters and just write something like "character x cannot be used"

ilesinge:
Having a look at the error message customization, thanks :)

felixroos:
one potential danger i see here is that some of these chars might become useful later to be used as new operators in mini notation. so it will become a breaking change if new characters are needed that are part of this set. many programming languages reserve a bunch of words that might become language features later, even if they are never used in the end.

ilesinge:
It's better with the customized error message, see below.

image

felixroos:
> It's better with the customized error message, see below.

ah perfect

ilesinge:
Regarding potential operator characters, it should be noted that the added characters do not include special characters, only letters that are not "universally" easily typeable since they are in other alphabets or include diacritics.
I'm not sure they make good candidates for operators.
We could still reduce the list though, but it would be to the detriment of some languages.

felixroos:
> Regarding potential operator characters, it should be noted that the added characters do not include special characters, only letters that are not "universally" easily typeable since they are in other alphabets or include diacritics. I'm not sure they make good candidates for operators. 

good point! so maybe we don't need to reserve any of those then

> We could still reduce the list though, but it would be to the detriment of some languages.

i think we can keep it as is then

felixroos:
maybe remove the 180 limit again then? More then 180 could be useful in some scenarios as some js errors are really long and might contain useful info at the end

ilesinge:
:heavy_check_mark: Removed the 180 limit

felixroos:
looks good, I think this is ready to merge then!

@yaxu this is pretty core / potentially tied to tidal compat so I'll ping you to ask if you have any objections?

yaxu:
Fine by me!

ilesinge:
Thanks!

PR #758, opened 2023-10-26
add play function

felixroos:
The play function allows adding a pattern to the scheduler, as an alternative to stacking it in the bottom:

```js
stack(
  s("bd"),
  s("~ cp")
)
```

can now also be written as

```js
s("bd").play()
s("~ cp").play()
```

this is also how it already works in https://github.com/tidalcycles/strudel/tree/main/packages/web

### DIFF
--- a/packages/core/repl.mjs
+++ b/packages/core/repl.mjs
@@ -24,6 +24,7 @@ export function repl({
     getTime,
     onToggle,
   });
+  let playPatterns = [];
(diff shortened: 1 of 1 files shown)


felixroos:
play now also returns the pattern itself, which opens up some interesting patterning possibilities:

```js
s("rim(3,8)")
  .pan(0).play()
  .pan(1).rev().play()
```

Issue #378, opened 2023-01-15
some < 7 note scales are broken..


felixroos:
e.g. `note("0 1 2 3 4 5 6 7 8 ".scale('D major pentatonic'))` will play `D3 E3 F#3 A3 B3 D3 E3 F#3 A3 B3` but should play `D3 E3 F#3 A3 B3 D4 E4 F#4 A4 B4` . so the octave does not count up. it works with `C major pentatonic`. the problem is the way that `scaleOffset` counts octaves. It will only count up / down when a note pitch class begins with 'C', so scales that do not include a 'C' based note will all be in the same octave. the algorithm should be changed (it is bad anyway, please do not read it :D)

felixroos:
started this in fix-scale-offset branch

PR #764, opened 2023-10-27
fix: scale offset

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/378

adds new function `scaleStep` as a replacement for `scaleOffset`. `scaleTranspose` still uses `scaleOffset`, but it might be removed in the future?


felixroos:
follow up: https://github.com/tidalcycles/strudel/issues/765

PR #767, opened 2023-10-28
Document adsr function

ilesinge:


felixroos:
and another one! thanks :)

PR #771, opened 2023-10-31
Add pick and squeeze functions

daslyfe:


felixroos:
thank you!

yaxu:
I know it's ported from tidal, but I think we might need to take back `squeeze` for something else in the future

felixroos:
> I know it's ported from tidal, but just flagging that we might need to take back `squeeze` for something else in the future

hm yeah it seems quite specific to receive that name that's also used in the alignment. maybe rename it right away?
maybe `pickSqueeze` ?

yaxu:
@felixroos Yes sounds good, ok with you @daslyfe ?

daslyfe:
> @felixroos Yes sounds good, ok with you @daslyfe ?

Yup! Another slightly shorter name that could work to would be pickFit

yaxu:
Hmm, `fit` is used for something else in tidal - a slightly obscure function for isorhythms. It is nice and short though and `squeeze` is a bit of a misnomer as the source cycle could be stretched to fit the target event, if it lasts longer than a cycle.. Hmm, naming things is hard!

In the interests of the perfect being the enemy of the good I think I favour keeping it as squeeze, quickly renaming this to `pickSqueeze`, and then looking into supporting `pick.squeeze`

jarmitage:
see also https://github.com/tidalcycles/Tidal/issues/1039

Issue #776, opened 2023-11-02
global repl aka more vanilla flavour


felixroos:
right now, the repl is wrapped in a react hook `useStrudel`, which complicates a lot of things.
It has basically no benefit, expect that it allows to run multiple repls at the same time with independent scheduling.

It would be dramatically simpler to just create a single `repl(...)` in vanilla land and react really only does the ui stuff.
Compare that to @strudel/web , where `repl` is not even used, just `webaudioScheduler`, though it misses some functionality, like setcps, fit, etc... maybe the web package could be refactored to use the same global `repl`.

This would of course also apply to the MiniRepl's where it probably leads to a strong performance boost in the docs.

TODO:

- [x] inject theme style automatically
- [x] fontFamily setting
- [x] draw
- [ ] what else?

felixroos:
most of this is fixed with https://github.com/tidalcycles/strudel/pull/876
although it does not use a global repl / each mini repl still has its own instance. but the vanillification has taken place

follow up: https://github.com/tidalcycles/strudel/issues/877

PR #779, opened 2023-11-02
vanilla repl

felixroos:
approaches https://github.com/tidalcycles/strudel/issues/776

hopeful outcomes:

- simplicity of the codebase (easier to read / change / maintain)
- better performance

this is a bigger refactoring, with the goal of getting rid of @strudel.cycles/react + its dependency on `@uiw/react-codemirror`. 
Currently, the main pieces of the react package are `useStrudel` (used in main repl) + `MiniRepl` (used in docs).
It also holds all of the codemirror themes.
I came to realize that react is not very practical to implement these things, as they are mostly data structures + state management without much ui. It was never a well thought out decision, as the use of react for these non-ui things just grew out of the ui written with react (which is still a good choice imo).
Instead, the plan is to implement the main repl + the mini repl with the @strudel/codemirror package, which is a much thinner wrapper around the core repl (repl.mjs). It is just vanilla js, using codemirror directly instead of going through another wrapper package (`@uiw/react-codemirror`).

What I've done so far:

- pull module loading / eval scope population / audio init into prebake + adapt Repl + MiniRepl to it
- improve codemirror package to support new highlighting logic + some settings (theme, font)
- manage repl state inside repl.mjs directly using nanostores (not sure about this yet)
- handle custom keyboard shortcuts in CodeMirror (react package) to see how good codemirror handles them + use them in MiniRepl + Repl
- refactor nano-repl
- smarter way to reset sounds on shuffle
- add /vanilla route with new version of main repl
- keeping / for testing + maybe PRs will split to be able to test it longer

todo:

- [x] implement remaining settings in codemirror.mjs
- [ ] test all main repl functions
- [x] refactor draw / onPaint logic
- [ ] refactor MiniRepl
- [ ] fix: `panic` will silence fx sends, then add reevaluate + panic stop
- [ ] test nano-repl


felixroos:
repackaged this partly in https://github.com/tidalcycles/strudel/pull/863

PR #784, opened 2023-11-05
Document pianoroll

ilesinge:
Add pianoroll, color and label function documentation for inline usage (reference tab or autocomplete)

Object key parameters were documented as classic parameters, for ease of formatting

### DIFF
--- a/packages/core/pianoroll.mjs
+++ b/packages/core/pianoroll.mjs
@@ -56,6 +56,40 @@ Pattern.prototype.pianoroll = function (options = {}) {
 
 // this function allows drawing a pianoroll without ties to Pattern.prototype
 // it will probably replace the above in the future
+
+/**
+ * Displays a midi-style piano roll
+ *
+ * @name pianoroll
+ * @param {Object} options Object containing all the optional following parameters as key value pairs:
+ * @param {integer} cycles defaults to 4 - number of cycles to be displayed at the same time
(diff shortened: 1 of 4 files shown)


packages/core/pianoroll.mjs
```diff
@@ -56,6 +56,40 @@ Pattern.prototype.pianoroll = function (options = {}) {
...
+
+/**
+ * Displays a midi-style piano roll
+ *
+ * @name pianoroll
+ * @param {Object} options Object containing all the optional following parameters as key value pairs:
+ * @param {integer} cycles defaults to 4 - number of cycles to be displayed at the same time
+ * @param {number} playhead 0 to 1, defaults to 0.5 - location of the active notes on the time axis
+ * @param {integer} vertical 0 (default) or 1 - displays the roll vertically
+ * @param {boolean} labels false (default) or true - displays labels on individual notes (see the label function)
```
felixroos, line 68: all booleans can either be 0 | 1 or false | true

ilesinge: I guess we can document them all as boolean : 1/0 ?

ilesinge:
@jarmitage would you like to review?

packages/core/pianoroll.mjs
felixroos: looks good

PR #798, opened 2023-11-08
Create phaser effect

daslyfe:


felixroos:
this sounds really good! what about adding a shortcut for the phaser controls? Like:

- ph for phaser
- phc for phasercenter
- phd for phaserdepth

felixroos:
another idea would be to be able to expose numStages and fOffset as params

daslyfe:
Cool, I can update it to add those additional params. Is there a preferred method for adding shortcuts for superdough?

felixroos:
I've added the synonyms + a doc entry.
Each string in the array of controls.mjs creates an alias. The first one is the one being set on the Hap.
You can also use an array for the first one, which will create a control that can set multiple keys at once using array / ":" notation.
Not entirely sure if we really need numStages and fOffset as params? Maybe it gets confusing if there are too many params?

daslyfe:
I didn’t know you could define shortcuts that way, it’s clever! I don’t necessarily think patterning stages or foffset changes the sound enough to be very useful so I think it’s probably fine as is. An idea I had for improving it in the future would be able to define different “models” based on different hardware phasers and set the parameters based on the model.

felixroos:
ok so just merge it as is?

daslyfe:
Yep sounds good

PR #800, opened 2023-11-11
Improve documentation for synonym functions

ilesinge:
- Adds a JSDoc plugin to better parse `@synonym` tags
- Show tooltips for synonym function names
- Show synonyms in API reference
- Simplify JSDoc React component

Question: should we move all JSDoc-related files from the root folder to a 'jsdoc' subfolder?


felixroos:
nice, didn't know you can write jsdoc plugins to do stuff like this. 

> Question: should we move all JSDoc-related files from the root folder to a 'jsdoc' subfolder?

sounds reasonable

ilesinge:
In my latest commits I moved jsdoc-related files to a jsdoc subfolder, except doc.json and undocumented.json.
I also changed the undocumented.json file so that path do not include the local path to the main project folder, only from the project root, so it can be updated by anyone.
I think the PR is ready :)

felixroos:
thanks! I've added some small additions:

- also show the synonyms in the Autocomplete component 
- there was a little bug where synonyms wouldn't work anymore on hover because entry.name was mutated. 
- all synonyms now work in the autocomplete feature as well

ilesinge:
Great, thank you 🙌

PR #801, opened 2023-11-11
Add doc for euclidLegatoRot, wordfall and slider

ilesinge:


packages/codemirror/slider.mjs
```diff
@@ -111,6 +111,15 @@ export const sliderPlugin = ViewPlugin.fromClass(
   },
 );
 
+/**
+ * Displays a slider widget to allow the user manipulate a value
+ *
+ * @name slider
+ * @param {number} value Initial value
+ * @param {number} min Minimum value
```
felixroos, line 119: optional, defaults to 0

packages/codemirror/slider.mjs
```diff
@@ -111,6 +111,15 @@ export const sliderPlugin = ViewPlugin.fromClass(
...
+ * @param {number} max Maximum value
```
felixroos, line 120: optional, defaults to 1

felixroos:
thanks! i think the optional params should be marked, rest looks good

PR #802, opened 2023-11-12
remove unwanted cm6 outline for strudelTheme

kasparsj:
Hi Felix, the outline is not visible by default (because of matching background color), but very obvious with a light background, especially when using hydra (see screenshots below).
May need to copy the fix to all other themes?

Screenshot 2023-11-13 at 00 06 28
Screenshot 2023-11-13 at 00 06 47


felixroos:
thanks! does it work if you just put the style in Repl.css instead?

kasparsj:
sure - changed :)

felixroos:
> sure - changed :)

thanks, now it's just the tests failing, you can fix it by running `pnpm codeformat` in the root folder

PR #805, opened 2023-11-16
tidal style d1 ... d9 functions + more

felixroos:
this is an attempt in adding tidal style d1 ... d9 functions + the p function and an additional q function, while still being backwards compatible with the old strudel style of not using any of these functions:

- `pPatterns` maps identifiers to patterns
- before each evaluation, `pPatterns` is cleared
- on evaluation, `.p(id)` sets the calling pattern on `pPatterns` using the given `id` as key
- after evaluation, if at least one key is found in `pPatterns`, the pattern used for playback is a stack of all `pPatterns`
- after evaluation, if `pPatterns` is empty, the old behavior is used, so the last expression is expected to be a pattern
- `.d1` ... `.d9` are just shortcuts for `.p(1)` ... `.p(9)`

so for example:

```js
s("bd").d1
s("hh*2").d2
```

will play both patterns stacked, as `pPatterns` will be `{1: s("bd"), 2: s("hh*2") }`.

```js
s("bd").d1
s("hh*2").d1
```

Here, `s("hh*2")` will win. 

Also, This still works:

```js
s("hh*2")
```

There is now also the `all` function:

```js
all(set(gain(.5)))
```

.. which just takes in a function that is applied to the evaluated pattern at the end.

+ I've also added a `shouldHush` flag that can be set on the `evaluate` function. 
If it is true, the evaluation will not clear the `pPatterns`, meaning you could do block based evaluation (with hidden state).

breaking changes:

There is now also a new top level hush function that will clear pPatterns and return silence (actually the hush function did exist before but it was just the non chained variant of `.hush()` which is not useful at all i think).

To support .d1 to .d9, i had to remove the transpiler feature that turns note variables into note strings, as it interfered with it. I don't think anyone really uses this feature, as mini notation is just more practical + it can also be confusing when you cannot name your variable after a note name..

### DIFF
--- a/packages/core/evaluate.mjs
+++ b/packages/core/evaluate.mjs
@@ -47,10 +47,5 @@ export const evaluate = async (code, transpiler) => {
   // if no transpiler is given, we expect a single instruction (!wrapExpression)
   const options = { wrapExpression: !!transpiler };
   let evaluated = await safeEval(code, options);
-  if (!isPattern(evaluated)) {
-    console.log('evaluated', evaluated);
-    const message = `got "${typeof evaluated}" instead of pattern`;
-    throw new Error(message + (typeof evaluated === 'function' ? ', did you forget to call a function?' : '.'));
-  }
   return { mode: 'javascript', pattern: evaluated, meta };
 };
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -1975,9 +1975,9 @@ export const press = register('press', function (pat) {
  *   s("hh*3")
  * )
  */
-export const hush = register('hush', function (pat) {
+Pattern.prototype.hush = function () {
   return silence;
-});
+};
 
 /**
  * Applies `rev` to a pattern every other cycle, so that the pattern alternates between forwards and backwards.
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -1178,7 +1178,7 @@ export function reify(thing) {
  * @return {Pattern}
  * @synonyms polyrhythm, pr
  * @example
- * stack(g3, b3, [e4, d4]).note() // "g3,b3,[e4,d4]".note()
+ * stack("g3", "b3", ["e4", "d4"]).note() // "g3,b3,[e4,d4]".note()
  */
 export function stack(...pats) {
   // Array test here is to avoid infinite recursions..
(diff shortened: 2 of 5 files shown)


yaxu:
Looks great! So 'd' is for dough?

In practice it's nice to be able to disable a pattern quickly, and maybe the quickest way would be to remove the 'd' target. I suppose this wouldn't work if it was the last and only pattern in the file. Maybe that'd be a bit confusing, but probably not too bad.

jarmitage:
I moved from `d` to `p` (for pattern) in my setup a while ago for three reasons:
- 'd' only made sense when targeting SuperDirt, so when using Tidal with anything else (or indeed with >1 OSCTarget at once) it felt incorrect
- semantically, `p` as pattern seemed closer to the reality, i.e. it's a pattern that is being sent/streamed to a thing (which might be a SomethingDirt), rather than a "dirt". Having a different letter for every target also doesn't seem meaningful.
- "what is `d`?" is a question that it very often asked by beginners in workshops, and I don't think it has a really satisfying answer beyond explaining some history of the project, which while valuable it kind of gets in the way of learning core concepts

felixroos:
> Looks great! So 'd' is for dough?

I guess so, as d in tidal is for dirt ? We could also add p1 .. p9 to be more close to p(1) ... p(9) edit: just read @jarmitage 's answer, and yes I also think pX makes more sense. 

> In practice it's nice to be able to disable a pattern quickly, and maybe the quickest way would be to remove the 'd' target. I suppose this wouldn't work if it was the last and only pattern in the file. Maybe that'd be a bit confusing, but probably not too bad.

For that reason (and for later code modification reasons) I added q1 .. q9 as well as q(x) which just return silence. so to make a pattern silent, just change the `d` or `p` to `q`, which coincidentally all are made of the same shape :) i wonder what b is for then..

felixroos:
for d vs p, we could also add both, and use the p variants in all examples and let the d variants be there as compatibility / variant for tidal people who prefer it

yaxu:
Yes I think it shouldn't be a big change from d to p, as that's just a 180 degree turn. Changing from p to q in a performance is easy as well, you can just flip it over.

felixroos:
``

> Yes I think it shouldn't be a big change from d to p, as that's just a 180 degree turn. Changing from p to q in a performance is easy as well, you can just flip it over.

hihi, maybe q will one day be used to send the pattern to the cue output...

jarmitage:
This is the cue output: `d-_-b`

yaxu:
Yes that would send to both headphones, `c-_-b` just to the left one

felixroos:
q-_-p is now in.. so i guess I'll just add it like it is and block eval / codemod is an experiment for the future based on this

PR #808, opened 2023-11-17
add options param to initHydra

kasparsj:
Another small one, but imho good to have. Allows to pass options to Hydra, and load personal forks of hydra-synth.


felixroos:
good one, thanks :)

PR #818, opened 2023-11-22
Hydra fixes and improvements

atfornes:
Fixing hydra integration:

  - Improve performace
  - Enable hydra audio detection 
  - Add example using hydra audio detection in documentation


felixroos:
thank you :) 
I've tried my best to merge main into this one, where https://github.com/tidalcycles/strudel/pull/808 was merged earlier.
I've also changed the canvas from absolute to fixed, so scrolling wont move the canvas. 
I am also curious why this fix works / why there were performance problems earlier, was it the way the canvas was initialized or was it because of `s0.init({ src: testCanvas })` ?

atfornes:
Thanks, the code is cleaner now :)

I think the performance issue was mainly due to the cloning of testCanvas: `const hydraCanvas = testCanvas.cloneNode(true);`

felixroos:
> Thanks, the code is cleaner now :)
> 
> I think the performance issue was mainly due to the cloning of testCanvas: `const hydraCanvas = testCanvas.cloneNode(true);`

ok good to know!

I've now also added a `feedStrudel` option to `initHydra` that will send the strudel canvas to s0:
 
Bildschirmfoto 2023-11-24 um 12 51 51

this is super fun to use, works with pianoroll, spiral and scope

PR #820, opened 2023-11-23
Multichannel audio

daslyfe:
This is an MVP for multichannel audio that supports routing to any output on the user's interface. examples: 

`note("e a b g").ch("2:3")`
`note("e a b g").ch(2)`
`note("e a b g").ch(saw.range(0,8).slow(4))`

enhancements that could be added in followup work would be: 

- setting the destination channels independently for orbits, which are currently being summed to 0&1 like normal
- support surround sound/multichannel audio files
- .q automatically routing to the cue channels
- advanced panning features such as panMulti
- allow to specify a specific output audio interface (not sure if this is possible)


felixroos:
just tested it with blackhole 16ch and it works great! I only wonder if the 0 indexing makes sense, as channel numbers are typically counted from 1. what do you think?

daslyfe:
> just tested it with blackhole 16ch and it works great! I only wonder if the 0 indexing makes sense, as channel numbers are typically counted from 1. what do you think?

I think that makes sense, I can update it early next week, or feel free to update it in the meantime

daslyfe:
This should be ready to go :) not sure why the hyperpop test is failing

felixroos:
> This should be ready to go, not sure why the hyperpop test is failing

some weird rounding error, i guess it doesn't matter

PR #838, opened 2023-12-04
Prevent 404 on Algolia crawls

ilesinge:
URLS starting by "./" are correctly interpreted by browsers but it makes Algolia crawler follow a bad link such as https://strudel.cc/learn/time-modifiers/functions/value-modifiers which results in a 404.


felixroos:
thanks! these might even break forks running with a /strudel base path

Issue #422, opened 2023-02-08
improve samples tab


felixroos:
- fix https://github.com/tidalcycles/strudel/issues/268
- show which samples are loaded in heap
- show which samples are cached in service worker (offline)
- can click samples to listen
- find a better way to group banks
- find a way to make all `n`s browsable

felixroos:
<img width="707" alt="image" src="https://github.com/user-attachments/assets/e880289f-10d3-4553-a35e-1e7f0364514a">

there could be a detail view for each sound, allowing you to preview all files in a folder / potentially see waveforms and active playheads

TodePond:
now that bankAlias has landed, it would be a good time to look at this again (note to self)

yaxu:
Could also consider licensing #1277

Issue #118, opened 2022-05-22
try loading samples via file system


felixroos:
using https://web.dev/file-system-access/ though it's chromium only, it might be really useful. For big sample banks, loading over the network is a trade off between availability and traffic

rabidmachine9:
Just wanted to give a small input here just in case it is not obvious to some people, you can serve your samples via a local server that has CORS enabled. 
The following server is based on python3  https://gist.github.com/rabidmachine9/2e06a5f872d53bdc1c03bb9c12f9df08 you just copy the file inside your "Samples" folder and start the server

`$ cd ~/Music/Samples`
`$ python python3-cors.py 8000`

this will start a local server on port 8000
then you just access your local files like that: https://strudel.tidalcycles.org/#c2FtcGxlcyh7CiAgbHNkOiAnRGlydC1TYW1wbGVzL21ldGFsLzAwMF8wLndhdicsCn0sICdodHRwOi8vbG9jYWxob3N0OjgwMDAvJyk7CnMoImxzZCBsc2QiKQo%3D

Hopefully this helps somebody!

felixroos:
hey thanks for the input! it might make sense to add a chapter to the docs to describe using a local sample server.
It's even possible to do this without needing to create a file and paste code, using something like [serve](https://www.npmjs.com/package/serve), running `npx serve ~/Music/Samples --cors` (requires nodejs). There are probably similar command line servers in the python sphere, which we should probably add for people that don't have node installed..

rabidmachine9:
Yes, this also can happen with and maybe it is nice to find an one liner for the job, the only reason I choose the python server script is because your can choose easier the domains you want to whitelist instead of just whitelisting all domains that the ----cors flag is doing( `response.setHeader('Access-Control-Allow-Origin', '*');` ). 
Whitelisting just a target domain probably makes things a little more secure

felixroos:
I am no security expert but I don't see how a local sample server with lax cors rules could pose a risk, as you can only access the files served in that folder, also only if you have access to localhost

bwagner:
Apparently, the simple [http server](https://docs.python.org/3/library/http.server.html#http-server-cli) in python does not support CORS. But there's an even [shorter script](https://stackoverflow.com/a/21957017/) that worked for me with @rabidmachine9's approach.

felixroos:
[...]

PR #839, opened 2023-12-05
Sound Import from local file system

daslyfe:
Adds the ability for the user to import samples from a local directory on their computer. 

Samples are written to an indexDB that persists until it is manually cleared.

Tested on firefox, safari, chrome, and tauri on MacOS

Screenshot 2023-12-04 at 7 07 07 PM


felixroos:
> clears after a session is closed, not sure if it is possible to maintain a selected directory upon reload due to obscuring of the filesystem

samples can potentially be cached in the indexdb, which is also what the pwa internally does

felixroos:
there was also this issue: https://github.com/tidalcycles/strudel/issues/118#issuecomment-1613757580

felixroos:
> Cool, I will try to see if I can set up the indexdb to cache the samples

made this a while ago: https://codesandbox.io/p/sandbox/indexdb-sample-cache-pie3rl

could be a good starting point

daslyfe:
@felixroos I got the indexdb working, your example was very helpful! I still want to clean up/comment the code and add a loading animation because it can take a minute when uploading hundreds of files.

daslyfe:

felixroos:
just tried on doesn't work on firefox for me.. i cannot select audio files when `directory / webkitdirectory` attributes are set + when I remove them, the sound is imported as undefined..

daslyfe:
> just tried on doesn't work on firefox for me.. i cannot select audio files when `directory / webkitdirectory` attributes are set + when I remove them, the sound is imported as undefined..

oh I think that is because it is set up to upload a whole directory of sounds, and then assign them indexed by sub folder. so if you have <sounds>/<drums>/<files> you would upload the entire "sounds" directory

daslyfe:
maybe it would be more clear if it said "import folder" or something like that

felixroos:
> > just tried on doesn't work on firefox for me.. i cannot select audio files when `directory / webkitdirectory` attributes are set + when I remove them, the sound is imported as undefined..
> 
> oh I think that is because it is set up to upload a whole directory of sounds, and then assign them indexed by sub folder (like in tidal). so if you have sounds/drums/files you would upload the entire "sounds" directory

ahh i see, so then it was just a misunderstanding on how this feature works. just tested with a directory an it worked!

daslyfe:
> > This should be ready to go now :)
> > The only thing I am unsure about is how I registered the blob with the registerSound function. It seems to work fine, but maybe it would load faster if I made another shorter url that pointed to the blob or something like that.
> 
> you mean a data url? no idea if that would make a difference or not. I'd say we just roll this out and test it a bit.
> 
> There should probably be some annotation to local sounds + the possibility to delete them in the future.
> 
> Maybe this can be done as part of a general overhaul of the sounds tab ref #422

Works for me!

Issue #841, opened 2023-12-05
refactor doc links to trailing slashes?


felixroos:
currently, all the doc links point to urls without trailing slash. the reason is this now closed issue: https://github.com/tidalcycles/strudel/issues/569 with that fixed, urls with trailing slash now work.

the algolia search results actually point to the versions with trailing slash. 
when clicking on a result, the sidebar anchor links will reload the whole page because they point to the slashless version (and sometimes they might not work, not sure when exactly).

also, the algolia crawler seems to first hit the slashless urls and then the redirect, which is why the crawler ignores many urls, polluting the output.

the slashless urls only work because, for directories, github pages redirects from URLs without trailing slash to the same URL with trailing slash as mentioned here: https://github.com/vite-pwa/astro/issues/23#issuecomment-1773392276 (although the service worker intercepts these redirects and serves the cached response somehow) + astro builds each page into its own directory by default (https://docs.astro.build/de/reference/configuration-reference/#buildformat defaults to 'directory')

TLDR; just refactor the links to get a cleaner algolia crawl + less flashing when clicking around the docs

felixroos:
I'll look into it

felixroos:
fixed in https://github.com/tidalcycles/strudel/pull/842

Issue #832, opened 2023-12-03
Some relative paths break when site base is not root ("/")


shiyouganai:
Hello,

Recently, I wrote my first strudel pattern and was intrigued by the display possibilities offered by the ["my patterns"](https://github.com/tidalcycles/strudel/tree/main/my-patterns) feature. I forked and deployed strudel on my own TLD but with a base path of `"/strudel"`, and unfortunately a handful of things broke which seem to have the root cause of using:
- Relative `href` paths in `.astro` file HTML
- Relative paths in calls to `samples()` in `prebake.mjs`

These relative paths do not seem to work "intuitively" under the astro framework when the site base is not `"/"` and will pop up with warnings in the development console phrased as:

`[dev] Requests for items in your public folder must also include your base.` (specific file information follows after)

You can see the breakage on the example deployment at https://felixroos.github.io/strudel/swatch/.

I hacked a fix for the immediately noticible broken functionality like drum sample packs and the swatch page, but many of the links in the `/workshop` pages are broken too. I do not know what a good way to fix those would be since working with these web frameworks is outside of my wheelhouse, and the workshop links are written in MDX files.

Here is a link to a diff of my [branch](https://github.com/tidalcycles/strudel/compare/main...shiyouganai:strudel:saved-patterns), and my deployed strudel based on that branch is here: https://shiyouganai.dev/strudel, with a now working swatch page at: https://shiyouganai.dev/strudel/swatch.

I would like to contribute the fix via PR in a more polished way which also addresses the MDX file links in the workshop pages, but I would like to ask for input about how best to go about it since I am very unfamiliar with MDX and how to get variable data like the site base into those links.

felixroos:
thanks for the detailed report! I think I've found a good way to fix all the links in MDX files, for which I've started this PR: https://github.com/tidalcycles/strudel/pull/842 also aiming to fix https://github.com/tidalcycles/strudel/issues/841 . 
The docs are a bit in flux at the moment because we are adding / improving the search. 
If you want to help out, you can open a PR to the branch I've created (slashocalypse) and add your existing fixes there!

felixroos:
with https://github.com/tidalcycles/strudel/pull/842 most of this is now fixed only the font urls are still broken.
For that, I've created this follow up issue: https://github.com/tidalcycles/strudel/issues/847

PR #842, opened 2023-12-05
improve slashing + base href behavior

felixroos:
+ prefix sidebar links & mdx links with base
+ add trailing slashes

- https://github.com/tidalcycles/strudel/issues/841
- https://github.com/tidalcycles/strudel/issues/832

### DIFF
--- a/website/astro.config.mjs
+++ b/website/astro.config.mjs
@@ -15,26 +15,32 @@ const site = `https://strudel.cc/`; // root url without a path
 const base = '/'; // base path of the strudel site
 const baseNoTrailing = base.endsWith('/') ? base.slice(0, -1) : base;
 
-// this rehype plugin converts relative anchor links to absolute ones
-// it wokrs by prepending the absolute page path to anchor links
-// example: #gain -> /learn/effects/#gain
+// this rehype plugin fixes relative links
+// it works by prepending the base + page path to anchor links
+// and by prepending the base path to other relative links starting with /
 // this is necessary when using a base href like <base href={base} />
-// in this setup, relative anchor links will always link to base, instead of the current page
-function absoluteAnchors() {
+// examples with base as "mybase":
+//   #gain -> /mybase/learn/effects/#gain
+//   /some/page -> /mybase/some/page
+function relativeURLFix() {
   return (tree, file) => {
     const chunks = file.history[0].split('/src/pages/'); // file.history[0] is the file path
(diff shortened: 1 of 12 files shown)


felixroos:
added a few more fixes

there is still a problem with the fonts, which are loaded from `index.css` e.g. `src: url('/fonts/teletext/EuropeanTeletext.ttf');`

not sure how to get the base href in there.. might work with an api endpoint

felixroos:
i guess I'll merge this as is, the font urls can be fixed in a separate PR. all the other problems should be fixed + it will hopefully give algolia a better next crawl

website/astro.config.mjs
```diff
@@ -13,24 +13,34 @@ import AstroPWA from '@vite-pwa/astro';
...
     return rehypeUrls((url) => {
-      if (!url.href.startsWith('#')) {
+      let newHref = baseNoTrailing;
+      if (url.href.startsWith('#')) {
+        // special case: a relative anchor link to the current page
+        newHref += `/${path}/${url.href}`;
+      } else if (url.href.startsWith('/')) {
+        // any other relative url starting with /
+        // NOTE: this does strip off serialized queries and fragments
+        newHref += url.pathname.endsWith('/') ? url.pathname : url.pathname + '/';
```
shiyouganai, line 37: I noticed today that using `url.pathname` here does cause at least one tiny regression since it strips fragments off the ends of relative links (which I didn't think was an issue at first): in this excerpt from docs
```
You might now be able to see this properly here: [open in REPL](/#YXdhaXQgaW5pdEh5ZHJhKCkKbGV0IHBhdHRlcm4gPSAiMyA0IDUgWzYgN10qMiIKc2hhcGUoSChwYXR0ZXJuKSkub3V0KG8wKQpuKHBhdHRlcm4pLnNjYWxlKCJBOm1pbm9yIikucGlhbm8oKS5yb29tKDEpIA%3D%3D)
```
 the fragment needs to stay. So I think quick rework of this to keep fragment is needed. I can do that.

shiyouganai (changes requested):

website/astro.config.mjs
felixroos: oh I haven't noticed that

shiyouganai: Hotfixed by https://github.com/tidalcycles/strudel/pull/845, thanks!

felixroos:
that did the trick

image

PR #854, opened 2023-12-08
Audio device selection

daslyfe:
Currently only Chromium browsers seem to have the ability to set the device ID for an audio context. clients without setSinkID will not show this dropdown.

Screenshot 2023-12-08 at 12 17 25 PM
Screenshot 2023-12-08 at 12 17 33 PM


felixroos:
just testing it, works great on my machine! some small points I've noticed:

- when I click on the dropdown, it asks if it can access the microphone. Might there be a way to not need to request for mic access? I'd expect this to be an output only selection so mic access is not really needed. I guess this comes from `await navigator.mediaDevices.getUserMedia({ audio: true });`[It seems enumerateDevices also works without this call](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices#examples)
- maybe label the dropdown "Audio Output Device" ?
- rename "-- select device --" to "System Standard" ?

daslyfe:
I believe the user needs to give permission in order to access the audio device (that is how I got it to work on my end), Im not sure if there is a way around it (maybe it depends on your system/ security config?). I don't think you can change the dialog message either even though it is a bit misleading (or if it is possible, I could not find a way to do it). The good news is that you should not keep hitting the dialog once you have approved it once. 

I updated the text based on your feedback :)

felixroos:
> I updated the text based on your feedback :)

thanks!

> I believe the user needs to give permission in order to access the audio device (that is how I got it to work on my end), Im not sure if there is a way around it (maybe it depends on your system/ security config?). I don't think you can change the dialog message either even though it is a bit misleading (or if it is possible, I could not find a way to do it). The good news is that you should not keep hitting the dialog once you have approved it once. It should also only trigger if the user clicks the button. It is a tiny annoyance, but hopefully not too bad.

ok i've now also tested it without `getUserMedia` and it only lists one unnamed output device. I guess this is just a browser thing that cannot be changed so it's fine.

One other thing that I've noticed is that once a device is chosen, the "System Standard" cannot be selected anymore. In many cases it might actually be desirable to use the System Standard, maybe there is a way to make it selectable. But we can also do that in a separate PR if you want.

daslyfe:
I could not figure out a way to get the id of the system default device, otherwise I agree it would be nice to select it again. Setting to null or “default” results in error

felixroos:
> I could not figure out a way to get the id of the system default device, otherwise I agree it would be nice to select it again. Setting to null or “default” results in error

ok maybe that's just another browser thing that has to be accepted if there's no other way.

Seems I've created some conflicts in settings.mjs :-/

felixroos:
maybe to clear the selection a reload could be requested?

daslyfe:
> maybe to clear the selection a reload could be requested?

I want to see if I can play around with it some more to find a workaround

daslyfe:
@felixroos I solved the issue, selecting "System Standard" resets the audio context and does not set the sinkID. This should be ready to go now.

daslyfe:
@felixroos I updated this to work with vanilla, it should be ready to go.

felixroos:
> @felixroos I updated this to work with vanilla, it should be ready to go.

thanks! the big changes are now all out of the door, so I can finally look at your PRs :)

felixroos:
just found a bug: when I switch between outputs, I sometimes get:

```
[cyclist] error: Failed to execute 'connect' on 'AudioNode': cannot connect to an AudioNode belonging to a different audio context.
```

@daslyfe any idea why this might happen?

PR #863, opened 2023-12-14
Vanilla repl 2

felixroos:
second take on https://github.com/tidalcycles/strudel/pull/779

this time keeping the existing repl as is and only duplicating logic to the codemirror package.
It should make sense to test this thoroughtly before replacing anything of the existing repl.

the good news: the editor (codemirror) with all its features now runs completely on vanilla js.


felixroos:
I've now added back the /vanilla route where the new vanilla implementation can be tested without interfering with the / route. I'd say this is ready to merge, the many lines are mostly the duplicated themes, the lockfile and the new examples (strudelmirror + /vanilla route). When this turns out fine, the next PR can replace the / route with the new logic + remove a LOT of code

Issue #522, opened 2023-03-17
better envelopes


felixroos:
envelopes will have pops when the hap is shorter than attack + decay. cancelAndHoldAtTime supposedly fixes this (in chrome only, see https://bugzilla.mozilla.org/show_bug.cgi?id=1308431 which is blocked by https://github.com/WebAudio/web-audio-api/issues/2437). it seems there is actually no real solution right now??
some clues could be found in one of:

- https://andremichelle.github.io/web-audio-api-automator/  (adsr timeConstant)
- https://github.com/velipso/adsrnode
- https://github.com/Tonejs/Tone.js/blob/aa58bf7fd1c045a35b819aa2e1f9e1bdf39dd204/Tone/component/envelope/Envelope.ts

felixroos:
another design that could be chosen is to just add a `t = Math.max(t, begin + attack + decay)` inside `stop` returned by `getEnvelope` to make sure the sound duration is at least completing the attack and decay phase (reaching sustain level) before starting the release. This should prevent all pops + might even be musically useful

felixroos:
fixed with https://github.com/tidalcycles/strudel/pull/861

PR #868, opened 2023-12-16
Further Envelope improvements

daslyfe:
I wanted to improve the envelope behavior to make them more intuitive and musically useful.

- current value is held at the end of the event time, and the release ramp begins at that value (like you would expect if you held down a note, and let it go, the release  would begin at that time from the current value)
- fixes glitchy envelope behavior for filters
- changes filter envelopes to use exponential curve (behavior is now more musically useful/ in line with most synthesizers due to how the ear perceives frequency)

- Default ADSR values are set based on which parameters are used to be more intuitive/less verbose for achieving desired effect. for example: sound(val).decay(val) will behave like a decay envelope

Test examples: 

`
// D
note("f a e c3").sound("gm_synth_strings_2:3").decay(0.1)
// A 
note("f a e c3").sound("gm_synth_strings_2:3").attack(0.1)
// AD
note("f a e c3").sound("gm_synth_strings_2:3").attack(0.2).decay(0.1)
// AR
note("f a e c3").sound("gm_synth_strings_2:3").attack(0.2).release(.7)
// ADSR 
note("f a e c3").sound("gm_synth_strings_2:3").attack(0.1).decay(0.1).sustain(.5).release(.3)

// filter example
note("c d d g f a".add(12)).sound("sawtooth").lpf(100).lpenv(18).lpattack(0.1).lpdecay(.2).lprelease(2).release(2)
`


Bubobubobubobubo:
This is super nice!

daslyfe:
@felixroos  There were quite a few edge cases caused by different behaviors between Firefox and chromium, but this should be ready to go now :)

daslyfe:
not sure why it is flagging prettier, it is passing locally

felixroos:
finally came around to look at this one!

I love the idea to change the envelope type based on which controls are set.
I've added 2 more functions `ad` + `ar` to more quickly use those types, so only one control is enough in these cases.
These functions allow both single values (value is used for both) or lists with 2 elements (set individual values):

```js
// AD
note("c a f e").ad(.1) // = .attack(.1).decay(.1)
note("c a f e").ad(".1:.2") // = .attack(.1).decay(.2)
// AR
note("c a f e").ar(.1) // = .attack(.1).release(.1)
note("c a f e").ar(".1:.2") // = .attack(.1).release(.2)
```

one regression i've noticed is that the fix from https://github.com/tidalcycles/strudel/pull/861 seems to have vanished. 
The problem was something like this would have produced a crack before above mentioned PR:

```js
note("c a f e").decay(1)
```

In the currently deployed version, the crack is gone, and instead, the decay plays till the end. 
When I do the same on this branch (envelope_improvements), I get cracks again.

I suspect the problem is somewhere [here](https://github.com/tidalcycles/strudel/pull/868/files#diff-098f4845f9eab40c0a81d0093eb9d59b7e17239bd749d8ff376fcc376d17878cR4-R31). 
iirc canceling a previously scheduled envelope is not really possible on firefox (as mentioned in https://github.com/tidalcycles/strudel/issues/522 )

as of my knowledge, the only 2 ways to avoid cracking envelopes on firefox is to either let it play till it's finished (as in https://github.com/tidalcycles/strudel/pull/861 ) or to implement a custom envelope using an AudioWorklet.

Letting it play till the end is not what a normal synth would do but I don't find it very unintuitive either..

What do you think?

felixroos:
regarding envelope canceling: one idea i haven't tested yet is to not cancel the param but just use another gain node for the release.. this would mean the envelope would continue during the release but it would probably sound more or less the same. don't know why it shouldn't work.. I think the `setTimeout` workaround is not really working

daslyfe:
This should be ready to go again

PR #869, opened 2023-12-21
Tree sitter: haskell parser

felixroos:
experiment on bringing a haskell parser into the browser


felixroos:
recreated with https://github.com/tidalcycles/strudel/pull/870 for less changes

PR #870, opened 2023-12-21
hs2js package / tidal parser

felixroos:
experiment on bringing a haskell parser into the browser

*edit from the future:* 

example usage:

```js
await initTidal()

tidal`
d1 
$ sub (note "12 0")
$ sometimes (|+ note "12")
$ jux rev $ voicing $ n "<0 5 4 2 3(3,8)/2>*8"
# chord "<Dm Dm7 Dm9 Dm11>"
# dec 0.5 # delay 0.5 # room 0.5 # vib "4:.25"
# crush 8 # s "sawtooth" # lpf 800 # lpd 0.1
# dist 1

d2 
$ s "RolandTR909_bd*4, hh(10,16), oh(-10,16)"
# clip (range 0.1 0.9 $ fast 5 $ saw)
# release 0.04 # room 0.5
`

all(x=>x._punchcard())
```


felixroos:
waiting for https://github.com/tree-sitter/tree-sitter-haskell/issues/108

felixroos:
here's a little haskell / tree-sitter playground https://github.com/felixroos/haskell-tree-sitter-playground/

felixroos:
> waiting for [tree-sitter/tree-sitter-haskell#108](https://github.com/tree-sitter/tree-sitter-haskell/issues/108)

it looks like this is not going to happen anytime soon. So to properly parse tidal style haskell, another parser has to be used :'(

yaxu:
Nice to see a draft PR with the support for `#` operators

felixroos:
> Nice to see a draft PR on tree-sitter-haskell with the support for `#` operators

yes, there is hope there is hope: https://github.com/tree-sitter/tree-sitter-haskell/pull/120

felixroos:
The general mechanism seems to be in place, now it's just a matter of implementing all missing operators and functions. Due to a non-existing type system + slight differences between strudel/superdough and tidal/superdirt we can't get a 100% match, but we can get close.
Maybe it makes sense to merge this as is, so people can play with it and give feedback?

yaxu:
I don't see why not!

felixroos:
for some reason, github actions doesn't understand symlinks, so when using the workspace version of hs2js, the postinstall task fails. For this reason I've now pinned hs2js on 0.0.8 (one version behind workspace), causing it to not create a symlink.. the version picked in the website package is for the wasm files anyway so it only has to be updated when they change. maybe a better solution can be found in the future: https://github.com/tidalcycles/strudel/issues/1106

PR #871, opened 2023-12-23
bugfix: sound select indexes out of bounds

daslyfe:
Round n to the nearest integer for selecting sounds, negative numbers will now select from the end of the sound array. 
Examples: 

sound("gm_electric_guitar_muted!4").n(saw.range(-2, 12).slow(2))

sound("bd!8").n(saw.range(0, 8).slow(1))


packages/soundfonts/fontloader.mjs
```diff
@@ -1,6 +1,7 @@
-import { noteToMidi, freqToMidi } from '@strudel.cycles/core';
+import { noteToMidi, freqToMidi, _mod } from '@strudel.cycles/core';
 import { getAudioContext, registerSound, getEnvelope } from '@strudel.cycles/webaudio';
 import gm from './gm.mjs';
+import { getSoundIndex } from '../superdough/util.mjs';
```
felixroos, line 4: iirc relative imports from another package might break stuff. to be safe, this should either be importing from `@strudel.cycles/superdough` (and make sure it's exported)

daslyfe: I couldn't get that import to work, so I just added the function to the utils in core

PR #874, opened 2023-12-26
Bug Fix #119: Clock drift

### LINKED ISSUE #119: [osc] mysterious clock drifting

daslyfe:
Updated cyclist based on @yaxu's comment https://github.com/tidalcycles/strudel/issues/119#issuecomment-1743687548 

all cps settings should now stay in time with other music programs/hardware.

clock should no longer accumulate rounding errors

### DIFF
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -11,17 +11,14 @@ export class Cyclist {
   constructor({ interval, onTrigger, onToggle, onError, getTime, latency = 0.1 }) {
     this.started = false;
     this.cps = 1;
-
-    // this.num_ticks_since_cps_change = 0;
+    this.num_ticks_since_cps_change = 0;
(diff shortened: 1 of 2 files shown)


felixroos:
very nice! the clock seems much more stable now. thanks so much!

yaxu:
Seconded, thanks a lot @daslyfe !

PR #882, opened 2023-12-31
 bugfix: suspend and close exisiting audio context when changing interface

daslyfe:
@felixroos do you still get that error you mentioned in https://github.com/tidalcycles/strudel/pull/854 with this update? I can't recreate it myself, but I have a hunch this might fix it.


felixroos:
seems fine now! thanks :)

PR #885, opened 2024-01-01
Showcase

felixroos:
adds new section "presentation" to docs where "Getting Started" has been moved and renamed "What is Strudel?" + a new page called "Showcase", where many strudel videos are presented in a random order.

not entirely sure if it's okay to just share them there without asking the people involved?


felixroos:
image

PR #900, opened 2024-01-11
REPL sync between windows

daslyfe:
using a shared worker,
- allows multiple instances to sync together
- clock no longer drops or slows down when browser is not on screen

- clock jitter on chrome is the same as before, still not sure what is going on there, might be unrelated to the scheduler


daslyfe:
FIxed the animation drops

felixroos:
the clock jitter is probably happening because the worker uses `performance.now()` (system clock) as a clock source instead of `AudioContext.currentTime` (audio clock). Sadly, an AudioContext cannot exist in a worker, so some other solution is needed to make sure the deadline / target time is accurate in the context of the audio clock

daslyfe:
@felixroos  this should be ready now, the clock is stabilized in firefox, and chrome behaves the same as before. Animations are working correctly. The clock no longer dies when window is unfocused. multiple windows can now sync together. I added a setCycles function as well, which we can integrate later.

daslyfe:
not sure what is going on with these test failures though...

felixroos:
> not sure what is going on with these test failures though...

self was not defined in node, I've wrapped it in a check

felixroos:
it's magic btw

felixroos:
I can review / test this in the coming days ok? on first sight it works great on my machine, I can also test it on my other machine then.

daslyfe:
> I can review / test this in the coming days ok? on first sight it works great on my machine, I can also test it on my other machine then.

Sounds good!

packages/core/clockworker.js
```diff
@@ -0,0 +1,128 @@
...
+  // the incoming port
+  const port = e.ports[0];
+  numPorts = numPorts + 1;
+  port.addEventListener('message', function (e) {
+    processMessage(e.data);
+  });
+  port.start(); // Required when using addEventListener. Otherwise called implicitly by onmessage setter.
+};
+
+function createClock(
```
felixroos, line 91: why has this been pulled out? i don't see any real difference between the old zyklus.mjs and this function, except some inputs being a closure now, also I don't understand why we need to round twice, once in getTime and once more in the while loop

packages/core/cyclist.mjs
```diff
@@ -1,93 +1,136 @@
...
+    let worker_time_dif = 0; // time difference between audio context clock and worker clock
+    let weight = 0; // the amount of weight that is applied to the current average when averaging a new time dif
+    const maxWeight = 400;
+    const precision = 10 ** 3; //round off time diff to prevent accumulating outliers
+
+    // the clock of the worker and the audio context clock can drift apart over time
+    // aditionally, the message time of the worker pinging the callback to process haps can be inconsistent.
+    // we need to keep a rolling weighted average of the time difference between the worker clock and audio context clock
+    // in order to schedule events consistently.
+    const setTimeReference = (time, workertime) => {
```
felixroos, line 34: this looks like something that could potentially be solved with https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/getOutputTimestamp

daslyfe: yes I think that could maybe help with the issue of audio context time being inconsistent if there are audio drops.

packages/core/clockworker.js
daslyfe: hmmm so I tried adding the import for zyklus, and it worked in chrome but not in firefox for me.

daslyfe: okay so it looks like importScripts works.It is supported for everything but deno it looks like https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts

packages/core/cyclist.mjs
daslyfe: After playing around with it, Im not sure how helpful have the performance time of the audio context, because we still need to schedule with a comparison between the worker clock relative to the audio ctx clock

daslyfe: maybe there is another way to solve this problem

daslyfe: I think they would be different because performance.now() on worker would be the active time since the worker was started, and the ctx performance time should be the time the ctx has been alive

felixroos:
I've added a `sync` flag to opt-in to the NeoCyclist, which is set by the Repl in the website. Flok also uses the repl so maybe set sync to false by default. When the next version is released, we can test if it also works in flok.

felixroos:
i'll try to fiddle around and see if there is a way to reuse zyklus, as it is identical to neozyklus

felixroos:
image

ah damnit

packages/core/cyclist.mjs
felixroos: there is performance.timeOrigin to know when the origin of performance.now(), so the worker could potentially be given its creation time.. but we can also tackle that in a future PR

felixroos:
we should test this on mobile before merging i think. mobile browsers often suck at supporting some apis, so might need to disable it on mobile if it doesn't work

daslyfe:
> we should test this on mobile before merging i think. mobile browsers often suck at supporting some apis, so might need to disable it on mobile if it doesn't work

It works on IPhone but not on Chrome on Android for some reason...

daslyfe:
@felixroos I added a condition to use cyclist if the client does not support shared workers and now android works again.

felixroos:
> @felixroos I added a condition to use cyclist if the client does not support shared workers and now android works again.

great! one more thing i've noticed is that with the neo cyclist, the time won't reset when you stop and start again, which might not always be desirable when you want to listen to the same place over and over again when writing. maybe we add a checkbox to allow switching sync off? Or maybe there is also a way to detect if all repls are stopped?

daslyfe:
> > @felixroos I added a condition to use cyclist if the client does not support shared workers and now android works again.
> 
> great! one more thing i've noticed is that with the neo cyclist, the time won't reset when you stop and start again, which might not always be desirable when you want to listen to the same place over and over again when writing. maybe we add a checkbox to allow switching sync off? Or maybe there is also a way to detect if all repls are stopped?

Good idea, I updated the behavior to stop the clock if all instances are stopped :) If a client disconnects while it is started would be the edge case I'm not sure how to solve. I don't think that is that much of a concern though

felixroos:
> Good idea, I updated the behavior to stop the clock if all instances are stopped :) If a client disconnects while it is started would be the edge case I'm not sure how to solve. I don't think that is that much of a concern though

cool! so maybe a checkbox to toggle sync on or off is not required then? 
I guess then this could be merged (finally). 
I've created https://github.com/tidalcycles/strudel/issues/988 as a follow up to potentially simplify the clock sync

daslyfe:
> > Good idea, I updated the behavior to stop the clock if all instances are stopped :) If a client disconnects while it is started would be the edge case I'm not sure how to solve. I don't think that is that much of a concern though
> 
> cool! so maybe a checkbox to toggle sync on or off is not required then? I guess then this could be merged (finally). I've created #988 as a follow up to potentially simplify the clock sync

sounds good! yes I hope there is a more straightforward way to simplify the this in the future. I will go ahead and merge :)

PR #903, opened 2024-01-12
fix: codemirror does not vanish anymore on hot reload

felixroos:
when saving certain files, for example superdough.mjs, the codemirror editor would just vanish.
It seems the hot reloading clears the #code element (Repl.jsx) without deleting the refs, so init won't be called again. 
Luckily, there is an API for vite to hook into hot reloading life cycle: https://vitejs.dev/guide/api-hmr#hot-dispose-cb

I managed to at least clear the refs and stop the previous repl, so it will be recreated when the #code ref fires again.
I am not sure if it's even possible to hot swap the editor without loosing the context. maybe it is but maybe it's not worth the hassle. @daslyfe any idea? how is this related to https://github.com/tidalcycles/strudel/pull/892 ?

### DIFF
--- a/packages/codemirror/codemirror.mjs
+++ b/packages/codemirror/codemirror.mjs
@@ -323,6 +323,7 @@ export class StrudelMirror {
     this.editor.dispatch({ changes });
   }
   clear() {
+    this.stop();
     this.onStartRepl && document.removeEventListener('start-repl', this.onStartRepl);
   }
 }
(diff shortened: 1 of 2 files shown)


daslyfe:
Sorry, forgot to respond to this one. iirc 892 was hot reloading failing to run on the newest pnpm version, not just the editor.

PR #904, opened 2024-01-12
add pregain + move postgain after sends

felixroos:
the `postgain` control was a bit misleading, as it was still applied before the effect sends. this PR adds another gain stage called `pregain` be the gain stage before the effect sends (but after local effects). Additionally, if a value is set to `postgain`, another gain stage will be added at the very end. this should allow full wet effects in the future. The signal path is now:

source -> gain -> local fx -> pregain -> sends -> postgain

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -87,7 +87,12 @@ const generic_params = [
    */
   ['gain'],
   /**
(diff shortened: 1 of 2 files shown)


daslyfe:
I wonder if it would be possible to evaluate the fx and chain them in the order they are written also, then you might not even need all three gain types.

PR #907, opened 2024-01-13
Add more vowel qualities for the vowels function

fnordomat:
In addition to the [a] [e] [i] [o] [u] that were already present, this adds

ae aa oe ue y uh un en an on

corresponding to IPA vowels
[æ] [ɑ] [ø] [y] [ɯ] [ʌ] [œ̃] [ɛ̃] [ɑ̃] [ɔ̃]

Formant values and gains were tuned empirically,
additionally validated by comparing with

- Korkmaz and Aytuǧ, Classification of Turkish Vowels Based on Formant Frequencies (http://dx.doi.org/10.1109/IDAP.2018.8620877)
- French Wikipedia: Formant (https://fr.wikipedia.org/wiki/Formant)
- A multitude of online sources for English vowels


yaxu:
Nice! As far as I know Unicode should work, if you'd like to add the symbols as additional aliases.

fnordomat:
I tried that and it worked, except for the nasal vowels, which the parser didn't like. But I can add the others!

yaxu:
Aha, well free to add them all and open an issue for the parsing, I'm looking at the parser at the moment

fnordomat:
OK, doing that now!

Hm, it's a bit unfortunate that I chose y for the high back unrounded vowel (Turkish ı), because according to IPA, this stands for high front rounded (ü). :thinking:

fnordomat:
Maybe like this, as a compromise:

```
get æ() { return this.ae; },
get ø() { return this.oe; },
get ɑ() { return this.aa; },
get å() { return this.aa; },
get ö() { return this.oe; },
get ü() { return this.ue; },
get ı() { return this.y; },
```

fnordomat:
Added an issue for the parser thing and a commit to define 7 aliases.

felixroos:
very nice! do we need to wait for https://github.com/tidalcycles/strudel/issues/908 or can this be merged before?

fnordomat:
I'd like to suggest that it's independent. Unless one wants to use strictly IPA symbols, the choice of aliases is arbitrary

PR #924, opened 2024-01-19
add pickF and pickmodF

geikha:
Allows to pick functions via a pattern of numbers, to apply to another pattern.

Example:
```javascript
note("<c2 d2>(3,8)").s("square")
    .pickF("<0 2> 1", [jux(rev),fast(2),x=>x.lpf(800)])
```

### DIFF
--- a/packages/core/signal.mjs
+++ b/packages/core/signal.mjs
@@ -214,6 +214,32 @@ export const pickmod = register('pickmod', function (lookup, pat) {
   return _pick(lookup, pat, true).innerJoin();
 });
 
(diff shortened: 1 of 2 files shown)


geikha:
I don't understand what's failing here. The examples work perfectly when I run them on the REPL!

felixroos:
yep that's normal, the examples automatically tested with snapshot testing. snapshots for new examples need to be generated by running `pnpm snapshot`. then push the updated .snap file and everything should be fine

felixroos:
thank youuu

PR #931, opened 2024-01-20
make 0.5hz cps the default

yaxu:
froos edit:
- breaking change: 0.5 cps is now default, giving more space in each pattern + being more similar to tidals default tempo
- adapted all doc pages + reference examples to make sense in the new default tempo
- patched database to add setcps(1) to all patterns that use the default tempo


felixroos:
@yaxu I've now adapted all doc pages from the bottom up, till audio effects, so I think I've met you from the other side..
also, I've swapped the db table from code to code_v1, where I've now patched all patterns to have a setcps(1) if no setcps (or similar) call is found. now at least all linked patterns should not break. the locally saved user patterns will still break tho

Issue #930, opened 2024-01-20
Pass cps in as state


yaxu:
Then `loopAt` and friends could be defined in `pattern.mjs` like:
```js
const _loopAt = function (factor, pat) {
  return new Pattern((state) =>
    pat.speed((1 / factor) * state.controls._cps)
    .unit('c')
    .slow(factor).query(state));
};
```

felixroos:
fixed with https://github.com/tidalcycles/strudel/pull/933

PR #932, opened 2024-01-20
Make splice cps-aware

yaxu:
Puts _cps in the pattern control state, as per #930

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -342,7 +342,7 @@ export class Pattern {
    */
   queryArc(begin, end, controls = {}) {
     try {
-      return this.query(new State(new TimeSpan(begin, end, controls)));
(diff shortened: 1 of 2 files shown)


packages/core/pattern.mjs
```diff
@@ -340,9 +340,9 @@ export class Pattern {
    * silence
    * @noAutocomplete
    */
-  queryArc(begin, end) {
+  queryArc(begin, end, controls = {}) {
     try {
-      return this.query(new State(new TimeSpan(begin, end)));
+      return this.query(new State(new TimeSpan(begin, end, controls)));
```
felixroos, line 345: shouldn't controls be passed to State and not TimeSpan?

yaxu: Oops yes, fixed.. It actually works now!

Issue #922, opened 2024-01-18
mini repl compatibility


felixroos:
some things don't work in the mini repl or are a bit buggy:

- functions using cps don't do anything (setcps etc..)
- sliders take up too much height
- what else?

If that's sorted out, the MiniRepl could be used for the Community Feed

felixroos:
cps fixable with https://github.com/tidalcycles/strudel/issues/930

PR #933, opened 2024-01-21
Refactor cps functions

felixroos:
following https://github.com/tidalcycles/strudel/pull/932 , the remaining functions that depend on scheduler cps ( `fit` / `loopAt` ) are now receiving the cps through pattern state, which gets the cps from the scheduler at query time.

Additionally, the functions that interact with the scheduler ( `all` , `hush` , `setcps` , `setcpm` ) did not work with multiple repls on the same page, because the last repl that was initialized got to set those functions on the eval scope. so any of those functions running in another repl would interact only with the last loaded scheduler. i fixed this by setting the functions on the eval scope right before evaluation, to make sure the Pattern is talking to the correct scheduler. This still feels a bit hacky, but it should work reliably (as long as 2 repls are not evaluated simultaneously)

this fixes the most annoying part of https://github.com/tidalcycles/strudel/issues/922 so I've also now added mini repls to the community bakery

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -427,7 +427,7 @@ export class Pattern {
    * @noAutocomplete
    */
   withHaps(func) {
-    return new Pattern((state) => func(this.query(state)));
+    return new Pattern((state) => func(this.query(state), state));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


felixroos:
fixes https://github.com/tidalcycles/strudel/issues/930

PR #936, opened 2024-01-23
fix midi issue on firefox and added quote error

Enelg52:
Fixed this error that happens on firefox 
```
[query]: can't access property 72, refs[input] is undefined
```
And added an error when you use double quotes on the midi device name. 

There is still a problem with the `WebMidi.enable()` function that never stops loading.

### DIFF
--- a/packages/midi/midi.mjs
+++ b/packages/midi/midi.mjs
@@ -167,10 +167,16 @@ let listeners = {};
 const refs = {};
 
 export async function midin(input) {
+  if (isPattern(input)) {
+    throw new Error(
+      `.midi does not accept Pattern input. Make sure to pass device name with single quotes. Example: .midi('${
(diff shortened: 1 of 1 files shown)


daslyfe:
I can't really think of a good reason why we should not allow patterning different midi devices, it could be cool.

felixroos:
> I can't really think of a good reason why we should not allow patterning different midi devices, it could be cool.

for `midin`, it's a bit more tricky because the function is not part of the main pattern so the querying would need to be done on the side. certainly doable but a little more involved. Patterning the input device is probably also not super interesting / useful? output device patterning seems more interesting + it's also easier, because the .midi function is in the main pattern. iirc the problem was that the function is async.

felixroos:
@Enelg52 thanks!

PR #941, opened 2024-01-26
Minor documentation error: Update first-sounds.mdx

mhetrick:
Update the tutorial text to reflect that the default cpm setting has changed from 60 to 30.


felixroos:
thanks, good catch!

PR #949, opened 2024-02-17
pickr and solo with stacking info

eefano:
Now stack()  (and the mininotation comma operator) populate the 'stacking' property of the hap context by pushing the stacking index into an array. The array handles nested stacking also. The outermost stacking level is the first element of the array.

The new solo() and mute() functions make use of this information by keeping only the haps that match (or not) the specified outermost stacking index. The new produced haps also have this index striped out from the stacking array, so subsequent calls of solo() can work with the inner stacking layers.

The implementation of solo() may be a little tricky. The original hap value is temporarily stored in context.tempvalue, modified to the stacking index, and then restored after the filtering operation is done.

The new pickr() and pickrmod() functions make use of solo(0), because restart("[1,2]") produces a pattern with duplicated haps. "[1,2]".solo(0) should be equivalent to "[1]", thus solving the problem.

I have also fixed two little typos in inhabit() and inhabitmod() declarations.

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -680,7 +680,7 @@ export class Pattern {
     const otherPat = reify(other);
     return otherPat.fmap((b) => this.fmap((a) => func(a)(b))).trigzeroJoin();
   }
-
+  
   //////////////////////////////////////////////////////////////////////
   // End-user methods.
   // Those beginning with an underscore (_) are 'patternified',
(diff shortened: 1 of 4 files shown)


felixroos:
I am not sure if it's a good idea to increase cost of the very fundamental stack function for this single use case, doing a lot of extra iterations and generating objects. A simpler version could be to add use tagging like this: https://strudel.cc/?R95e2cPMjyBH

eefano:
No problem! I went down the rabbit hole trying to make pickr() work as intended, and came up with this overkill idea. I definitely learned something in the process!

PR #956, opened 2024-02-22
'Enable Bracket Matching' option in Codemirror

eefano:
The matching brackets will be underlined with the cursor color.


daslyfe:
This looks great! @felixroos  maybe this should be enabled by default?

felixroos:
very cool, also that it includes mini notation brackets!

> This looks great! @felixroos maybe this should be enabled by default?

yep, makes sense

PR #958, opened 2024-02-25
add synonyms for euclid functions 

daslyfe:
euc = euclid
eucl = euclidLegato
eucr = euclidRot
euclr = euclidLegatoRot

### DIFF
--- a/packages/core/euclid.mjs
+++ b/packages/core/euclid.mjs
@@ -10,9 +10,8 @@ https://rohandrape.net/?t=hmt
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more details. You should have received a copy of the GNU Affero General Public License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import { Pattern, timeCat, register, silence } from './pattern.mjs';
+import { timeCat, register, silence } from './pattern.mjs';
 import { rotate, flatten, splitAt, zipWith } from './util.mjs';
-import Fraction from './fraction.mjs';
 
 const left = function (n, x) {
   const [ons, offs] = n;
@@ -58,6 +57,7 @@ export const bjork = function (ons, steps) {
  *
  * @memberof Pattern
  * @name euclid
+ * @synonyms euc
  * @param {number} pulses the number of onsets / beats
  * @param {number} steps the number of steps to fill
  * @returns Pattern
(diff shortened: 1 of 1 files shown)


felixroos:
maybe it makes sense to allow the first param of euclid to be an array, so ":" notation can be used, allowing the last param to be optional:

```js
s("bd").euclid("3:8")
// is the same as
s("bd").euclid(3,8)
// and
s("bd").euclid("3:8:1")
// is the same as
s("bd").euclidRot(3,8,1)
```

this would eliminate the need to use / learn euclidRot. Legato could be achieved with the second param being negative:

```js
s("bd").euclid("3:-8")
// is the same as
s("bd").euclidLegato(3,8)
// and
s("bd").euclid("3:-8:1")
// is the same as
s("bd").euclidLegatoRot(3,-8,1)
```

we can still merge these shortcuts, but I think having a more flexible `euclid` function makes the other ones more or less obsolete

daslyfe:
> maybe it makes sense to allow the first param of euclid to be an array, so ":" notation can be used, allowing the last param to be optional:
> 
> ```js
> s("bd").euclid("3:8")
> // is the same as
> s("bd").euclid(3,8)
> // and
> s("bd").euclid("3:8:1")
> // is the same as
> s("bd").euclidRot(3,8,1)
> ```
> 
> this would eliminate the need to use / learn euclidRot. Legato could be achieved with the second param being negative:
> 
> ```js
> s("bd").euclid("3:-8")
> // is the same as
> s("bd").euclidLegato(3,8)
> // and
> s("bd").euclid("3:-8:1")
> // is the same as
> s("bd").euclidLegatoRot(3,-8,1)
> ```
> 
> we can still merge these shortcuts, but I think having a more flexible `euclid` function makes the other ones more or less obsolete. edit: I generally find myself using euclid with structs more often:
> 
> ```js
> s("bd")
> .struct("x(3,8) x*2")
> ```
> 
> this would also benefit if a negative second param would make the thing legato

I really like the idea of using the array pattern syntax with negativve numbers. Maybe the second array param being rotate though would be more useful? unless you are doing really weird stuff with rhythms. here is another thought: 

```js
s("bd").euclid("5:2 -3:1", 8)```

the first half of the cycle would be 5 with a rotation of 2, the second half would be 3 with a rotation of 1, but backwards. This way you could get a huge variety of rhythms with just a few numbers in a pattern.  I think having a second argument for the division makes sense because it will probably stay stagnant for most use cases?

**Comment by felixroos** [comment]
> Maybe the second array param being rotate though would be more useful

I'd rather have the array params be the same order as the regular params, otherwise it can get confusing.. you could still do:

```js
s("bd").euclid("5:8:2 -3:8:1")
// or
s("bd").euclid("[5 -3]:8:[2 1]")
```

Using an array for the first param is generally a good way to be able to call a function with optional args (which is not possible with curried functions otherwise). If the argument structure of the array notation is the same as for the regular function, you don't have to remember more stuff

daslyfe:
Hmm I see what you mean. tbh Ive never really played much with the euclid mininotation, it seems much more powerful. It could be cool also do be able to do ```x([5:2 -3:1], 8)``` 
to easily get a wider variety of patterns (where the second param is rotate).

Issue #30, opened 2022-04-09
add missing mini notation features


felixroos:
Compared to mainline tidal, the following features are missing from the mini notation:

- [x] Tie symbols "_" https://github.com/tidalcycles/strudel/pull/926
- [x] feet marking "." https://github.com/tidalcycles/strudel/pull/926
- [x] standalone "!" https://github.com/tidalcycles/strudel/pull/926
- [x] random choice "|" https://github.com/tidalcycles/strudel/pull/165
- [x] Random removal "?" https://github.com/tidalcycles/strudel/pull/165
- [x] Polymetric sequences "{ ... }" https://github.com/tidalcycles/strudel/pull/336
- [x] Fixed steps using "%"
- [ ] https://github.com/tidalcycles/strudel/issues/65
- [x] https://github.com/tidalcycles/strudel/issues/72
- [x] https://github.com/tidalcycles/strudel/issues/176
- [x] colons 
- [x] `c3(3,8)?` does not work (`[c4(3,8)]?` does)
- [x] polyrhythms with <> e.g. `<bd hh, ht lt mt>` https://github.com/tidalcycles/strudel/pull/886
- [x] range operator "0 .. 6"` https://github.com/tidalcycles/strudel/pull/716

For most of them to actually work, the corresponding functions need to be implemented in core.

felixroos:
correlates with https://github.com/tidalcycles/strudel/issues/31

filippoguida:
Tie PR https://github.com/tidalcycles/strudel/pull/47/

filippoguida:
Feet Marking PR https://github.com/tidalcycles/strudel/pull/49

felixroos:
related https://github.com/tidalcycles/strudel/issues/72

bpow:
Related #165 for random choice and random removal.

bpow:
Are the various "word-like" operators towards the bottom part of the krill.pegjs file desired?

Because PEG parsers don't perform backtracking, I think if you were to pass something to the parser like `mini('slow 2 $ "a b"')` that "slow" and "2" both end up being recognized as "slices" (because slices match before the individual command operators) and then the parser gets to the '$' and doesn't know what to do.

[...]

PR #959, opened 2024-02-25
feat: can now invert euclid pulses with negative numbers

felixroos:
the `bjork` function now inverts the output if the first param is negative. this is useful to create pairs of euclid patterns that fill each others gaps:

```js
s("bd(3,8), hh(-3,8)")
```

iirc tidal also has this feature


felixroos:
ref https://github.com/tidalcycles/strudel/issues/30

Issue #686, opened 2023-08-31
refactor controls


felixroos:
- add `registerControl` function, similar to `register` to be able to export controls directly (not wrapped into controls object)
- allow more **control** over controls, for example being able to opt-out of "interpret array as sequence"

PR #962, opened 2024-02-27
controls refactoring: simplify exports

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/686

exports all controls directly instead of throwing them into the `controls` objects, removing the need to import controls explicitly all the time

this change should not break existing setups as long as everything is imported from the core index.mjs and not directly from the controls.mjs


felixroos:
i think it makes sense to merge this one now and think more deeply about grouping with https://github.com/tidalcycles/strudel/issues/964

PR #966, opened 2024-02-28
fix for transpose(): preserve hap value object structure

eefano:
Now transpose() and scaleTranspose() can be called without disrupting the hap value object contents, it detects the object type and alters the note property only.

Now this function can be safely used in place of add()/sub(), in cases when control properties are already set inside the hap and we don't want to get altered by add().

Use case:
```
n("0 1 2").pan(0.0).gain(0.5).scale("c:major").transpose(12)
```


felixroos:
thanks! while testing i've found some other weird things for which I've created this issue: https://github.com/tidalcycles/strudel/issues/969 

I remember having added transpose to be able to stay in note land when rendering to sheet music at some point, being able to get correct accidentals.. but I've not done that so far

PR #968, opened 2024-02-29
add debounce to logger

felixroos:
when a runtime error occurs, the console is spammed with logs at 20Hz (query time). this can sometimes even block the ui. to avoid that, I've added a throttling / debounce to the logger, where the same message can only be repeated once per second.


felixroos:
ref https://github.com/tidalcycles/strudel/pull/968

PR #973, opened 2024-02-29
Nested controls

felixroos:
poc for https://github.com/tidalcycles/strudel/issues/288

allows doing things like

```js
s("bd".pan(0))
```


felixroos:
started by @eefano here: https://github.com/tidalcycles/strudel/pull/972

PR #975, opened 2024-03-04
Added ability to mute a synth/sound after it was triggered

klibardi:
I would like to introduce an ability to control synth or sample after it was triggered. For now I just need the ability to mute, but I am open to implementing control over other aspects of sound too.

Here is an example of what I'm using this for:

```
await WebMidi.enable()
const myInput = WebMidi.inputs[0];
const noteControls = Array.from(Array(128)).map(x => null)
myInput.removeListener('noteon');
myInput.removeListener('noteoff');
myInput.addListener('noteon', e => {
  const [n, vel] = e.dataBytes;
  superdough({ note: n, s: 'piano', postgain: vel / 127 }, 0.05, 10).then(function (x) {
    noteControls[n] = x;
  });
});
myInput.addListener('noteoff', e => {
  const [note] = e.dataBytes;
  const { mute } = noteControls[note]
  mute(0.1);
  noteControls[note] = null
})

sound("metal casio casio casio")
```

### DIFF
--- a/packages/superdough/superdough.mjs
+++ b/packages/superdough/superdough.mjs
@@ -518,6 +518,7 @@ export const superdough = async (value, deadline, hapDuration) => {
   // toDisconnect = all the node that should be disconnected in onended callback
   // this is crucial for performance
   toDisconnect = chain.concat([delaySend, reverbSend, analyserSend]);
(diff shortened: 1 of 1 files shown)


klibardi:
Just a link for reference of what I'd like to accomplish in the long run:
https://club.tidalcycles.org/t/baked-in-support-for-midi-in-audio-in-and-audio-devices-channels/5098

PR #976, opened 2024-03-04
Beat-oriented functionality

yaxu:
This PR allows `timeCat("a b", "c d e")` to mean the same as `"a b c d e"`, and `polymeter("a b", "c d e")` the same as `"{a b, c d e}"` via the `weight` property of patterns.

There is also a new `__pure` property, so that 'pure' patterns can be treated as values by some functions.

Plus a new experimental `beatCat` function, which is similar to `timeCat`, but if an argument is a list, the sequence will repeat once for each element in the list.


yaxu:
This is towards #987

PR #977, opened 2024-03-04
Little fix for control.mjs + Proposed modification to scale()

eefano:
EDIT for clarity:
This patch addresses two separate instances:
- A little bug fix in controls.mjs 
- A proposed change to scale() to make these two lines equivalent in behaviour:
```
"0 c#".scale("c:major").note()...[chain of control functions]   // legacy version
"0 c#"...[chain of control functions]...scale("c:major")  // proposed version
```
The proposed change allows to things:
- being able to express a mixed pattern of number and notes, like the legacy version
- being able to put the scale() function at the very end of a structured stack() of instruments, all 'tuned' on the same scale, avoiding the repetition of the call for every pattern line in the stack.

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -15,7 +15,7 @@ export function createParam(names) {
     let bag;
     // check if we have an object with an unnamed control (.value)
     if (typeof xs === 'object' && xs.value !== undefined) {
-      bag = { ...xs}; // grab props that are already there
+      bag = { ...xs }; // grab props that are already there
       xs = xs.value; // grab the unnamed control for this one
       delete bag.value;
     }
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -15,7 +15,7 @@ export function createParam(names) {
     let bag;
     // check if we have an object with an unnamed control (.value)
     if (typeof xs === 'object' && xs.value !== undefined) {
-      bag = xs; // grab props that are already there
+      bag = { ...xs}; // grab props that are already there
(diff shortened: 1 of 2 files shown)


eefano:
I have modified the scale() function also, to handle this case:
```
n("0 a").scale("c:major")
```
that was causing:
```expected hap.value to be an object, but got "a"```

felixroos:
> Test case:
> 
> ```
> "a5".gain(0.5).note()
> ```

i don't understand what the regression is? this didn't work before https://github.com/tidalcycles/strudel/pull/973 and now it works. On the live site it still doesn't work because it has not been deployed yet, but it works locally.

> n("0 a").scale("c:major")

Should this be a valid use case? I agree that the error is certainly confusing, but do you think the `a` should be treated as a note here?

eefano:
To be fair, i'm experiencing the problem on the live site only, not in the actual main branch, so my fix on control is not really relevant to the problem (but it still prevents the altering of the original hap by the delete statement).

eefano:
Found the problematic commit:
https://github.com/eefano/strudel/commit/5c71bb95a0ec1384b45a0716d9fb2963036dea3d
When legato was "de-legacyfied", it followed the behaviour of the other control functions.

eefano:
Ok with the latest modification, the object's value property is transformed by scale(), but only when n: is not already present.
At least something like that will be allowed:
```"0 c#".pan(0).scale("c:major").note()```

The n() variants will behave as usual.

felixroos:
can you update the issue's top post to summarize what these changes are doing?

PR #980, opened 2024-03-05
little fix for withVal

eefano:
the variable bag was a reference to the original object, the fix maintains immutability


felixroos:
thanks!

PR #989, opened 2024-03-14
inline viz / widgets package

felixroos:
adds new inline viz widgets:

- adds `._pianoroll` to render an inline pianoroll (name might be changed later)
- adds `._scope` to render an inline `.scope`(name might be changed later9
- (adds `._spiral` to render an inline spiral (name might be changed later) (commented out for now)

image


felixroos:
this one seems ready

Issue #990, opened 2024-03-17
transpile label expressions to `p` calls


felixroos:
label expressions could be used to set the key of the pattern:

```js
p1: s("bd*4")

p2: note("c e g b")
```

could be made equivalent to:

```js
s("bd*4").p('p1')
note("c e g b").p('p2')
```

label expressions are a niche js syntax that is very much unused, so perfect for overloading.
This benefits readability, especially in combination with inline widgets:

before:

<img width="552" alt="image" src="https://github.com/tidalcycles/strudel/assets/12023032/c3bcbe31-889b-4cd2-a679-0827ce3f5dd8">

after:

<img width="539" alt="image" src="https://github.com/tidalcycles/strudel/assets/12023032/d78f467b-a83d-4382-ba7b-58c1610efd29">

strickinato:
I've been exploring using strudel with coffeescript. 

I have a preference for it's syntax, I think it's nice not to mess around with the parentheses while code.

eg.

```coffeescript
n "0 3 <5 7> 3"
  .scale "c:minor"
  .sound "piano"
  .jux rev

s "bd <sd cp>"
```

Interestingly, coffeescript doesn't support labelled expressions! niche indeed!

Which means that the above has to be written as:

```coffeescript
n "0 3 <5 7> 3"
  .scale "c:minor"
  .sound "piano"
  .jux rev
  .p1

s "bd <sd cp>"
  .p2
```

Not trying to push an agenda, just putting this note here for posterity. <3

felixroos:
> I've been exploring using strudel with coffeescript.
> 
> I have a preference for it's syntax, I think it's nice not to mess around with the parentheses while code.
> 
> eg.
> 
> n "0 3 <5 7> 3"
>   .scale "c:minor"
>   .sound "piano"
>   .jux rev
> 
> s "bd <sd cp>"
> 
> Interestingly, coffeescript doesn't support labelled expressions! niche indeed!
> 
> Which means that the above has to be written as:
> 
> n "0 3 <5 7> 3"
>   .scale "c:minor"
>   .sound "piano"
>   .jux rev
>   .p1
> 
> s "bd <sd cp>"
>   .p2
> 
> Not trying to push an agenda, just putting this note here for posterity - and calling out that this decision has an extremely mild negative impact on strudel-as-a-transpilation-target 😄
> 
> Maybe this will matter one day, maybe not 🤷
> 
[...]

PR #991, opened 2024-03-17
Labeled statements

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/990

with this, inline vizualisations per pattern now look like this:

image


felixroos:
with this, inline scopes per track now look like this:

image

PR #1000, opened 2024-03-19
accidentals in scale degrees

eefano:
now this is supported:
```
"0 1 1b 1bb 1# 1## -2#".scale("c4:major").note().slow(4)
```


felixroos:
1000 !!! thanks :)

Issue #51, opened 2022-04-12
Patternable tempo changes via `cps`


yaxu:
There is `cpm` in core but it assumes a scheduler running at 1 cps.

It would be great if `cps` could be patternable, e.g.
```
cps(saw.range(0.9,1.8).slow(16))
```
or
```
sound("bd [bd bd]").cps(1,1.5)
```

felixroos:
I wonder if the assumption of 1 cycle / second has any disadvantage? If not, we could implement cps like cpm

yaxu:
Hm, I think having a global cycle rate that can be changed is pretty important. Personally I don't really like 1cps, the default in tidal is `0.5625`.

felixroos:
But anything against implementing it with slow, assuming a hard coded cps? I kind of like 1s because it's normalized

yaxu:
`slow` doesn't hold state, so `slow (saw.segment(8).range(1,2))` would miss a lot of events.

With something like: `"a b".cps(1.5).every(3,rev)`

The `rev` would not work on cycle boundaries.

More generally, when it comes to syncing between systems, it would be difficult to do that without a global clock.

felixroos:
makes sense. so we need a global clock that dictates how fast the scheduler will run. also thought syncing would be great. `.cps` would then have a side effect to talk to the scheduler.

yaxu:
SuperDirt likes to have some metadata - the current `cps`, the duration (as `delta`) of the event in seconds, and the event onset (wholeorpart.begin) in cycles (as `cycle`). So it could be good for the scheduler to provide this info (or a way to get it) to the callback.

felixroos:
i made some improvements to the scheduler + sketched cps option [DEMO](https://raw.githack.com/tidalcycles/strudel/86a68d78c13a3160a39b981a299fff83527c44ee/packages/webaudio/cps-sketch/index.html#c2VxdWVuY2UoMSwyKS5tdWwoNTUvMikgLy8gZnJlcXVlbmNpZXMKLm11bChzbG93Y2F0KDEsMikpCi5tdWwoc2xvd2NhdCgxLDMvMiw0LzMsNS8zKS5zbG93KDgpKQouZmFzdCgzKQouZnJlcSgpCi52ZWxvY2l0eSguNSkKLnMoJ3Nhd3Rvb3RoJykKLmN1dG9mZig4MDApCi5vdXQoKQo%3D)

[...]

PR #1001, opened 2024-03-19
Make cps patternable

eefano:
Little hack, you can change the global inner clock with the cps() control function. Very dangerous but very satisfying. You can bring the cps to 0 and stop the playback, also. Negative numbers work in theory (the pianoroll goes in reverse) but no correct audio is produced.

Practical example:

https://youtu.be/No1dQWJKQNU


felixroos:
I don't think this is a hack.. this partly fixes https://github.com/tidalcycles/strudel/issues/51

You could also use this like `s("bd hh").cps(saw.slow(4).range(.1,1))` which looks less hacky.
To make this complete, I'd say 2 things are missing:

- the .cpm function needs to use .cps internally, so patterning .cpm also behaves like expected
- setcps and friends also need to work with a pattern

+ when https://github.com/tidalcycles/strudel/pull/1004 is merged, this one probably needs to be updated to use this.setCps..

yaxu:
Yes tidal has this functionality too.

Based on my experience with tidal, there could be problems when the cps changes part way through a frame. This can result in events getting doubled up or missed.

yaxu:
I posted a summary of the situation here: https://github.com/tidalcycles/strudel/issues/51#issuecomment-2478814512

packages/core/cyclist.mjs
```diff
@@ -52,6 +52,10 @@ export class Cyclist {
               // see https://github.com/tidalcycles/strudel/pull/1004
               const deadline = targetTime - phase;
               onTrigger?.(hap, deadline, duration, this.cps, targetTime);
+              if (hap.value.cps !== undefined && this.cps != hap.value.cps) {
+                this.cps = hap.value.cps;
+                this.num_ticks_since_cps_change = 0;
```
yaxu, line 57: I suppose this shouldn't be 0 as we're partway through a tick. Perhaps somewhere between -1 and 0 ??

eefano: It's an integer counter...  the tick is the smallest unit of handled time. In a tick you feed a fixed-size audio buffer. You could go down to the single sample but i think performance could be killed.

yaxu:
Ok merged, thanks!

PR #1015, opened 2024-03-24
Document signals

ilesinge:
I also suggest that we add a dedicated "Random modifiers" documentation page, separated from https://strudel.cc/learn/signals/

What do you think ?


felixroos:
oh nice, that's a good chunk of missing doc! having a separate page for the random stuff sounds reasonable.

you can also run `report-undocumented` to see how the stuff you've added vanishes from `undocumented.json` :)

ilesinge:
I just updated the PR with a minor documentation fix, random modifiers on their own page, and undocumented.json cleaning !

felixroos:
> I just updated the PR with a minor documentation fix, random modifiers on their own page, and undocumented.json cleaning !

thanks!

Issue #1026, opened 2024-03-29
Remove dangerous arithmetic


yaxu:
Arithmetic is dangerous knowledge and should be stopped.

Also, currently `n("0 1 2").gain(2).add(12)` results in gain of `14` (as well as n of `"12 13 14"`), which probably isn't what was meant.

Possible solutions (some not mutually exclusive):

1. Don't support adding numbers to values at all. It could be a no-op, or it could raise an error (although we don't have good error handling)
2. Add shorthands for `add(n(12))` and friends.. Maybe `add.n` is possible.
3. Try to be helpful by only applying such things to particular controls, e.g. `n` and `note`
4. Add some clamping to 'dangerous' controls. Adding e.g. `overgain` control if you really want to go over the limit.
5. Keep as-is

Note that `n("3").fmap(x => x + 3).firstCycleValues` results in a pattern with the value `"[object Object]3"`. I guess that's probably fine left like that.

felixroos:
> Arithmetic is dangerous knowledge and should be stopped.

The secret knowledge should be buried before it gets in the wrong hands

some thoughts:

1. throwing an error only if it can be done at eval time (to avoid silence) which is not possible i think. so no-op might be best
2. that would save a lot of typing + increase readability a lot
3. maybe a bit too magical? i'd prefer 2 if possible
4. yep

felixroos:
https://github.com/tidalcycles/strudel/pull/1030 this one does a no-op and logs a warning to the console at runtime

PR #1030, opened 2024-03-29
remove dangerous arithmetic feature

felixroos:
fixes https://github.com/tidalcycles/strudel/issues/1026

currently just logs `[warn]: dangerous arithmetic` when something like `n(1).add(1)` is attempted.
Or maybe remove the log?


yaxu:
Heh maybe a less mysterious warning is better.. "Can't do arithmetic on control pattern." ?

PR #1040, opened 2024-04-01
WIP: make soundfont base url configurable

fnordomat:


packages/soundfonts/fontloader.mjs
```diff
@@ -9,14 +10,28 @@ import {
 } from '@strudel/webaudio';
 import gm from './gm.mjs';
 
+export const defaultFontloaderConfig = {
+  soundfontUrl: 'https://felixroos.github.io/webaudiofontdata/sound'
+}
+
+export const fontloaderConfigMap = persistentMap('strudel-config-fontloader', defaultFontloaderConfig);
```
felixroos, line 17: not sure if a persistentMap is best. wouldn't a simple global variable be enough? it would make sure the default fallback is used in the next session, so there is no hidden state stored in the browser. The fact that an alternative url is used would be visible in the code at all times

fnordomat: you're right, I'll change that.

felixroos:
thanks! i guess this one's ready now?

felixroos:
@fnordomat ok to merge?

felixroos:
this seems to work! allows using soundfonts offline, merging now..

PR #1053, opened 2024-04-11
Workshop edits

felixroos:
- add drum image
- introduce `<...>` earlier, because it's weird to introduce cpm as the only way to make a regular sequence slower.
- I've also tried to replace .cpm with setcps for changing tempo but it seems the mini repl does some strange things sometimes with setcps so i've kept it for now.


yaxu:
Yep I think `.cpm` is a bit confusing and it'd be better not included in the tutorial, and maybe deprecated.

felixroos:
> Yep I think `.cpm` is a bit confusing and it'd be better not included in the tutorial, and maybe deprecated.

I think it might still be good to introduce `setcpm`, because the number range is more spaced out compared to `setcps` + it's easier to translate to bpm

PR #1059, opened 2024-04-17
anonymous patterns + muting

felixroos:
1. adds the ability to create an anonymous patterns with a `$:` label

```js
$: s("bd")
$: s("hh*4")
```

this is handy because you won't have to name your patterns. I found myself just keyboard mashing random characters in a live situation because i just needed a unique name quickly.

2. adds the ability to mute a labeled pattern by adding a _ before or after:

```js
p1_: s("bd")
$_: s("hh*4")
```

I've started out with just a prefix, but I've noticed that the click area of the first column in each line is half as big as the click aread of the other columns. Maybe just a suffix is enough..

credits to @geikha for the idea

### DIFF
--- a/packages/core/repl.mjs
+++ b/packages/core/repl.mjs
@@ -54,10 +54,12 @@ export function repl({
   const scheduler =
     sync && typeof SharedWorker != 'undefined' ? new NeoCyclist(schedulerOptions) : new Cyclist(schedulerOptions);
   let pPatterns = {};
+  let anonymousIndex = 0;
   let allTransform;
 
   const hush = function () {
     pPatterns = {};
+    anonymousIndex = 0;
     allTransform = undefined;
     return silence;
   };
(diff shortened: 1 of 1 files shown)


yaxu:
Looks handy!

Would it be possible to allow any label to be used more than once like this?

felixroos:
> Looks handy!
> 
> Would it be possible to allow any label to be used more than once like this?

rn, labels overwrite each other (last one wins), which would be useful for block based evaluation in tidal style. The `evaluate` function of the repl already has a [shouldHush flag](https://github.com/tidalcycles/strudel/blob/93d151b943872f2c8f14dd51ea6aca55b828fdcb/packages/core/repl.mjs#L135), which currently always returns true, but maybe the editor will have a setting someday to evaluate only the block where the cursor is at..

Another alternative to block based evaluation would be to add an editor feature that mutes all labels with the same name on eval:

```js
p1: s("bd")

_p1: s("bd(3,8)")
``` 

if we'd eval the second pattern, it could become:

```js
_p1: s("bd")

p1: s("bd(3,8)")
``` 

this would have the same benefit of being able to switch between patterns quickly, but without hidden state. So it would mean that the state of the code reflects the state of evaluation, which is desirable imo

PR #1074, opened 2024-04-27
clarify `off` in pattern-effects.mdx

cleary:
I had a bit of difficulty understanding `off` as it was presented, so I've attempted to simplify it:
1. take out the `delay` effect which is obscuring things (since we're using a delayed `off` pattern)
2. add a more complete explanation of the syntax


daslyfe:
the example is broken because the // needs to be removed before .off, can you update it?

felixroos:
a bit late to the party.. i've fixed the linting problem. thanks cleary for clarification :)

PR #1078, opened 2024-04-29
Chop chop

yaxu:
Improved `chop`, that respects existing begin/end controls (e.g. from a previous chop).


felixroos:
nice one. I've been thinking this could be nice to have for slice as well, e.g. first cutting file to loop with begin / end, then slicing the loop instead of the whole file

yaxu:
Yes also `striate`, which I think is extra useful as doing a series of striates doesn't just multiply up the slices as with chop.

PR #1084, opened 2024-05-02
Improve performance of ! (replicate)

yaxu:
Re-implement with weighted repeatCycles

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -2310,7 +2310,13 @@ export const { iterBack, iterback } = register(
 export const { repeatCycles } = register(
   'repeatCycles',
   function (n, pat) {
-    return slowcat(...Array(n).fill(pat));
+    return new Pattern(function (state) {
+      const cycle = state.span.begin.sam();
+      const source_cycle = cycle.div(n).sam();
+      const delta = cycle.sub(source_cycle);
+      state = state.withSpan((span) => span.withTime((spant) => spant.sub(delta)));
+      return pat.query(state).map((hap) => hap.withSpan((span) => span.withTime((spant) => spant.add(delta))));
+    }).splitQueries();
   },
   true,
   true,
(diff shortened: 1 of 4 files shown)


yaxu:
* `s("<bd(3,8) -!50000>")` is fine now
* `s("<bd(3,8) -!50000>")` takes a while to start, but then runs fine (although the kicks never play because it took so long to get started)
* `s("<bd(3,8) -!500000>")` says too many function arguments for some reason??

yaxu:
Ah, `repeatCycles` itself is just `return slowcat(...Array(n).fill(pat));`.

felixroos:
> Ah, `repeatCycles` itself is just `return slowcat(...Array(n).fill(pat));`.

that is probably because `repeatCycles` uses `slowcat(...Array(n).fill(pat))` . maybe there is a way to implement this without each element being one argument?

felixroos:
> Ah, `repeatCycles` itself is just `return slowcat(...Array(n).fill(pat));`.

could that be implemented with a step based function like `s_cat`? maybe something like: https://strudel.cc/?0Ni_ayF8qs_Y

yaxu:
`s("<bd(3,8) -!500000000000000000>")` now works fine :)

felixroos:
> `s("<bd(3,8) -!500000000000000000>")` now works fine :)

great!

yaxu:
The s_cat solution is interesting but results in missed/duplicated haps..
https://strudel.cc/?cID8aKangF3_

yaxu:
The 9th kick is doubled up for `s("<cp  bd(3,8)!500000000000000000>")` :/

yaxu:
I'll merge for now though!

felixroos:
> The 9th kick is doubled up for `s("<cp bd(3,8)!500000000000000000>")` :/

hmm maybe this only happens with very very large numbers due to some overflow or float precision problems?

yaxu:
> > The 9th kick is doubled up for `s("<cp bd(3,8)!500000000000000000>")` :/
> 
> hmm maybe this only happens with very very large numbers due to some overflow or float precision problems?

It should all be rationals so shouldn't be precision.

500000000000000000 is about 8000000000 years though so not a huge issue for the next 200000000 generations of people

PR #161, opened 2022-07-19
Strudel CLI REPL

felixroos:
```
  ███████╗████████╗██████╗ ██╗   ██╗██████╗ ███████╗██╗     
  ██╔════╝╚══██╔══╝██╔══██╗██║   ██║██╔══██╗██╔════╝██║     
  ███████╗   ██║   ██████╔╝██║   ██║██║  ██║█████╗  ██║     
  ╚════██║   ██║   ██╔══██╗██║   ██║██║  ██║██╔══╝  ██║     
  ███████║   ██║   ██║  ██║╚██████╔╝██████╔╝███████╗███████╗
  ╚══════╝   ╚═╝   ╚═╝  ╚═╝ ╚═════╝ ╚═════╝ ╚══════╝╚══════╝


welcome to the experimental strudel cli <3

available samples: bd (1), hh (1), oh (1), old (7), sd (2), snap (1)   
you can add samples to the samples folder (.wav, .ogg, .mp3)

please enter a pattern, e.g. s("bd sd,hh*6")

>>>
```

This PR is a proof of concept for a Strudel REPL that runs in a terminal via node.

- uses nodes builtin [repl](https://nodejs.org/api/repl.html) (there are probably fancier things to use, just used it for simplicity)
- uses [Node Web Audio API](https://www.npmjs.com/package/web-audio-api) (alpha) + [node-speaker](https://github.com/TooTallNate/node-speaker)
- automatically loads all samples inside the sample folder

Future Ideas:

- add osc to have a way to run `strudel` -> `superdirt` with terminal only
- expose an API to have a strudel backend that can be controlled via HTTP requests
- package as single executable via [pkg](https://github.com/vercel/pkg) (like osc server)
- try using [elementary](https://www.elementary.audio/docs/packages/node-renderer) instead of node web audio api

I won't continue this path for now, just wanted to try out the idea. 
This PR is just a reminder that this is a thing that could exist

Usage Instructions:

```sh
# checkout cli-repl branch
cd server && npm i
node ./repl.js
```

PR #1088, opened 2024-05-03
node + superdirt experiment

felixroos:
a little experiment for a minimal strudel node runtime. the pattern is evaluated by saving a file and the haps are sent to via osc to superdirt.


felixroos:
this is a newer version of https://github.com/tidalcycles/strudel/pull/161
memo to self: add ascii logo

PR #1089, opened 2024-05-06
fix: missing events due to premature worklet cleanup

felixroos:
worklets could get cleaned up before the input was connected, resulting in silent events


daslyfe (commented):
Tested, works great, activity monitor stays steady as well

PR #1091, opened 2024-05-07
Use sessionStorage for viewingPatternData and activePattern

### LINKED ISSUE #1049: It's easy to overwrite user patterns by mistake
Reproduced in firefox and chrome (linux):

1. create a new pattern (patterns -> user -> new), paste in this awesome pattern:
  ```javascript
  // "awesome pattern" @by yaxu
  s("bd sd")
  ```
3. run the pattern
4. Open another pattern in a different tab, e.g.: https://strudel.cc/?PUqxlzDp7SSW
5. run the pattern

Awesome pattern has been overwritten with terrible pattern!

kasparsj:
localStorage, which is shared between tabs, was being used for these, and although it wasn't [live] synchronised between tabs, it would e.g. restore to the last saved data (accross all tabs) upon refresh/reload/new window. 

sessionStorage is separate for each tab, but persists tab data across refreshes.

this should fix #1049 as it's tracking the data about which pattern is active/selected for each tab separately.

### DIFF
--- a/website/src/user_pattern_utils.mjs
+++ b/website/src/user_pattern_utils.mjs
@@ -21,39 +21,25 @@ export const patternFilterName = {
   user: 'user',
 };
 
-export let sessionAtom = persistentAtom;
+const sessionAtom = (name, initial = undefined) => {
+  const storage = typeof sessionStorage !== 'undefined' ? sessionStorage : {};
+  const store = atom(typeof storage[name] !== 'undefined' ? storage[name] : initial);
+  store.listen((newValue) => {
+    if (typeof newValue === 'undefined') {
+      delete storage[name];
+    } else {
+      storage[name] = newValue;
+    }
+  });
+  return store;
+};
 
-if (typeof sessionStorage !== 'undefined') {
-  const identity = (a) => a;
-  sessionAtom = (name, initial = undefined, opts = {}) => {
-    let encode = opts.encode || identity;
-    let decode = opts.decode || identity;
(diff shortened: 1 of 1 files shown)


felixroos:
i meant this:

```js
const initialActivePattern = sessionStorage.getItem('activePattern');
const $activePattern = atom('activePattern', initialActivePattern);
$activePattern.listen((value) => sessionStorage.setItem('activePattern', value));
```

does that make sense?

kasparsj:
I see, my idea was to keep the api same as `persistentAtom` and base the implementation on this (no idea why it wasn't merged):
https://github.com/nanostores/persistent/pull/30

felixroos:
> I see, my idea was to keep the api same as persistentAtom and base the implementation on this :
https://github.com/nanostores/persistent/pull/30 (no idea why it wasn't merged)

`atom` is also a `Store` type (`WritableAtom`) which works for the part of the API that we use (`set`, `get` and passing it to `useStore`) so I don't see why we need more complexity here

felixroos:
thanks, this now looks good. In the future `viewingPatternData` might be refactored a little bit to not need parsing on every render (PatternsTab > parseJSON), but I guess that can be done in a separate PR. Also I wonder if a browser that doesn't support sessionStorage can even run the rest of strudel :P

yaxu:
Is this ready to merge?

felixroos:
> Is this ready to merge?

yup! thanks @kasparsj

PR #1094, opened 2024-05-08
add Pattern.every to the docs?

kasparsj:
Coming from TidalCycles I had to dig into the code to find that `every` has been renamed to `firstOf`. I think better to have it clear in the docs.

Now I see it appears in the Reference tab, but initially I was search trough the Learn section and not finding.

### DIFF
--- a/website/src/pages/learn/conditional-modifiers.mdx
+++ b/website/src/pages/learn/conditional-modifiers.mdx
@@ -18,7 +18,7 @@ import { JsDoc } from '../../docs/JsDoc';
 
 ## every
 
-<JsDoc client:idle name="Pattern.every" h={0}  />
+<JsDoc client:idle name="Pattern.every" h={0} />
 
 ## when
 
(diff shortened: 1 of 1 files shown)


yaxu:
Looks good to me, is it still a draft?

felixroos:
every is an alias of `firstOf`, so maybe add it using the `@synonyms` jsdoc tag?

kasparsj:
> every is an alias of `firstOf`, so maybe add it using the `@synonyms` jsdoc tag?

synonyms don't seem to show up in search though

felixroos:
> > every is an alias of `firstOf`, so maybe add it using the `@synonyms` jsdoc tag?
> 
> synonyms don't seem to show up in search though

indeed, i didn't know that. so then we need to fix that: https://github.com/tidalcycles/strudel/issues/1099
adding synonyms as separate headings would mean a lot of duplication / redundancy

kasparsj:
ok, good, closing this in favour of #1099

PR #1095, opened 2024-05-10
Gain Modulation with calculated modulation per event

daslyfe:
This pattern could be used as the basis for all kinds of interesting modulations.

An example: 
`$: note("{f g c d}%16".add("{12 7 0}%16")).am("<4 2 16 8>").amskew("<.5 1 0>").s("sawtooth")`

The timing of the modulation for the non-synced clock is a little off, and I haven't been able to figure out why yet. Maybe @felixroos you might have an idea? The synced clock timing is accurate.

### DIFF
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -56,7 +56,7 @@ export class Cyclist {
               // the following line is dumb and only here for backwards compatibility
               // see https://github.com/tidalcycles/strudel/pull/1004
               const deadline = targetTime - phase;
-              onTrigger?.(hap, deadline, duration, this.cps, targetTime);
+              onTrigger?.(hap, deadline, duration, this.cps, targetTime, end);
             }
           });
         } catch (e) {
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -34,6 +34,8 @@ export class Cyclist {
 
         try {
           const begin = this.lastEnd;
+          const cycle = this.now();
+
           this.lastBegin = begin;
           const end = this.num_cycles_at_cps_change + num_cycles_since_cps_change;
           this.lastEnd = end;
@@ -56,7 +58,7 @@ export class Cyclist {
               // the following line is dumb and only here for backwards compatibility
               // see https://github.com/tidalcycles/strudel/pull/1004
               const deadline = targetTime - phase;
-              onTrigger?.(hap, deadline, duration, this.cps, targetTime, end);
+              onTrigger?.(hap, deadline, duration, this.cps, targetTime, cycle);
             }
           });
         } catch (e) {
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -34,12 +34,15 @@ export class Cyclist {
 
         try {
           const begin = this.lastEnd;
-          const cycle = this.now();
 
           this.lastBegin = begin;
           const end = this.num_cycles_at_cps_change + num_cycles_since_cps_change;
+
           this.lastEnd = end;
+
           this.lastTick = phase;
+
+          const cycle = this.now();
 
           if (phase < t) {
             // avoid querying haps that are in the past anyway
--- a/packages/core/neocyclist.mjs
+++ b/packages/core/neocyclist.mjs
@@ -4,6 +4,7 @@ Copyright (C) 2022 Strudel contributors - see <https://github.com/tidalcycles/st
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more details. You should have received a copy of the GNU Affero General Public License along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
+import Fraction from 'fraction.js';
 import { logger } from './logger.mjs';
 
 export class NeoCyclist {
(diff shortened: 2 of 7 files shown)


felixroos:
just a quick idea: maybe `tremolo` should be an alias of `am` ? (Then also `pm` should be an alias of `vibrato`)
need to dig into the rest of it soon..

daslyfe:
> just a quick idea: maybe `tremolo` should be an alias of `am` ? (Then also `pm` should be an alias of `vibrato`) need to dig into the rest of it soon..

Yeah that makes sense to me 👍

daslyfe:
Okay this should be ready to go now. The cycle calculation jitters a bit on the non synced chromium clock but isn't very noticeableand can probably be improved with better math.

daslyfe:
I updated the non synced clock to match the synced clock in order to make the cps calculation more accurate on chrome. I also brought over the setCycle function that can be integrated in new features :)

felixroos:
I've tested it and it works great! 
It would be good to find a syntax to allow switching between absolute phase (like am) and relative phase (like all others rn). I imagine being able to apply tremolo with a relative/resetting phase could also be something desirable. It probably makes sense to make relative phase the default and add some keyword for absolute phase. Maybe something like `am.lfo()`.
It also depends on which other effects might be suitable for absolute phase. Could this also be used for non repeating effects like envelopes?

packages/core/cyclist.mjs
```diff
@@ -8,61 +8,51 @@ import createClock from './zyklus.mjs';
...
+    this.interval = interval;
+
     this.clock = createClock(
       getTime,
       // called slightly before each cycle
-      (phase, duration, _, t) => {
-        if (this.num_ticks_since_cps_change === 0) {
-          this.num_cycles_at_cps_change = this.lastEnd;
-          this.seconds_at_cps_change = phase;
+      (phase, duration, _, time) => {
```
felixroos, line 28: What's the idea behind these changes? I can't quite figure out the intention by just looking at the code o_O

packages/superdough/superdough.mjs
```diff
@@ -260,7 +260,7 @@ export function resetGlobalEffects() {
   analysersData = {};
 }
 
-export const superdough = async (value, t, hapDuration) => {
+export const superdough = async (value, t, hapDuration, cps, cycle) => {
```
felixroos, line 263: could there be a default value for cycle to make am (and potential future friends) be relative to seconds? so am(4) would be 4Hz. thinking about using superdough without strudel, where cycle is a non existing concept

felixroos:
> I've tested it and it works great! It would be good to find a syntax to allow switching between absolute phase (like am) and relative phase (like all others rn). I imagine being able to apply tremolo with a relative/resetting phase could also be something desirable. It probably makes sense to make relative phase the default and add some keyword for absolute phase. Maybe something like `am.lfo()`. It also depends on which other effects might be suitable for absolute phase. Could this also be used for non repeating effects like envelopes?

the switching between relative and absolute is probably not needed yet in this PR, just an idea for the future (tho it implies a breaking change)

packages/core/cyclist.mjs
daslyfe: The cycle calculation has to be adjusted slightly to compensate for when the event is actually played so that the phase of the modulation can be lined up. I already had made the cycle calculation on the neocyclist, so I just brought that logic over as well as the setCycle function. I could not figure out how to get the cycle offset adjustment to calculate properly with current cyclist implementation. Maybe you might know a better way to get this value?

packages/core/neocyclist.mjs
```diff
@@ -86,7 +83,7 @@ export class NeoCyclist {
             this.latency +
             this.worker_time_dif;
           const duration = hap.duration / this.cps;
-          onTrigger?.(hap, 0, duration, this.cps, targetTime);
+          onTrigger?.(hap, 0, duration, this.cps, targetTime, this.cycle);
```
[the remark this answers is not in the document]
daslyfe: That was the original plan, but the events don't line up perfectly unless you get the adjusted cycle that accounts for latency etc. Honestly could use some help here, maybe there is a better way to do this.

packages/superdough/superdough.mjs
daslyfe: yeah you could just pass in the time in seconds since clock start and it should work that way. I agree having both options would be nice for different effects

felixroos:
> > just a quick idea: maybe `tremolo` should be an alias of `am` ? (Then also `pm` should be an alias of `vibrato`) need to dig into the rest of it soon..
> 
> Yeah that makes sense to me 👍

thinking again about this, I realized there is a distinction between `fm` and `vib`:

- `fm` is relative to the hap frequency, which allows keeping the timbre of different notes the same (fmh controls the ratio of modulator and carrier frequency)
- `vib` uses an absolute modulator frequency, independent of the carrier frequency (typically used for lower frequency modulation)

just from a naming perspective, it might make sense for `am` to be frequency relative as well? The frequency independent version would probably be `tremolo` then? Combining this with absolute / relative phase, we have 4 variants:

|                            | absolute frequency | relative frequency |
| --------------- | --------------------| ------------------| 
| relative phase   | vib / tremolo             | fm / am                   | 
| absolute phase | vib.sync ? / tremolo.sync ? |  fm.sync ? / am.sync ? |

not sure if relative frequency + absolute phase is even needed..

daslyfe:
> > > just a quick idea: maybe `tremolo` should be an alias of `am` ? (Then also `pm` should be an alias of `vibrato`) need to dig into the rest of it soon..
> > 
> > 
> > Yeah that makes sense to me 👍
> 
> thinking again about this, I realized there is a distinction between `fm` and `vib`:
> 
> * `fm` is relative to the hap frequency, which allows keeping the timbre of different notes the same (fmh controls the ratio of modulator and carrier frequency)
> * `vib` uses an absolute modulator frequency, independent of the carrier frequency (typically used for lower frequency modulation)
> 
> just from a naming perspective, it might make sense for `am` to be frequency relative as well? The frequency independent version would probably be `tremolo` then? Combining this with absolute / relative phase, we have 4 variants:
> 
> absolute frequency	 relative frequency
> relative phase	vib / tremolo	fm / am
> absolute phase	vib.sync ? / tremolo.sync ?	fm.sync ? / am.sync ?
> not sure if relative frequency + absolute phase is even needed..

I think having different words for cycle based speed and frequency based might limit us when other modulations are added that don’t have similar word equivalents. I think it would be better to have something like amBasis(“<frequency cycle”) with cycle always being the default because it is much more rhythmically useful

PR #1102, opened 2024-05-16
Samples tab improvements

daslyfe:
-Allows users to select individual files on browsers that do not support folder uploads (safari and ios). This can be expanded upon later to allow uploading either/or 

-samples are now sorted alphabetically 
-import sounds button wraps to next line on too-small screens instead of getting cropped out


felixroos:
looks good!

PR #1103, opened 2024-05-18
Add analog-style ladder filter

daslyfe:
adapted from https://github.com/TheBouteillacBear/webaudioworklet-wasm?tab=MIT-1-ov-file

example pattern: 
note("c f g g a c d4").fast(2)
 .sound('sawtooth')
 .lpf(200).fanchor(0)
 .lpenv(4).lpq(1).drive("<.5 0 2>")
 .ftype("ladder")


felixroos:
sounds super nice! I've noticed the lpq range is different from the other filter types. I think it should be somewhat equal to avoid surprises when switching types

daslyfe:
> sounds super nice! I've noticed the lpq range is different from the other filter types. I think it should be somewhat equal to avoid surprises when switching types

cool I updated it to be as similar as I could with the different response curves

packages/superdough/superdough.mjs
```diff
@@ -288,24 +284,25 @@ export const superdough = async (value, t, hapDuration) => {
...
+    drive = 0.69,
     // low pass
     cutoff,
     lpenv,
     lpattack,
     lpdecay,
     lpsustain,
     lprelease,
-    resonance = 1,
+    resonance = 0.5,
```
felixroos, line 297: is this still needed after taming the resonance curve? would be a breaking change

test/__snapshots__/shared.test.mjs.snap
felixroos, line 1: not sure if we still need this one.. probably not

packages/superdough/superdough.mjs
daslyfe: updated the default back to normal :)

PR #1104, opened 2024-05-18
fix little dub tune example

lukad:
The dub tune example from the worshop's effects page is missing a `.` before the `room` effect.
dub-tune


felixroos:
thanks for spotting! that must have gone missing at some point :)

PR #1110, opened 2024-05-23
Calculate phaser modulation phase based on time

daslyfe:
phaser no longer has to rely on orbits and free running lfos, because the phase of the modulator is calculated based on the time. the phaser should sound exactly the same as before

```
$: note("{f g g c d a a#}%8")
  .gain(1).s("sawtooth").lpenv(4).lpf(200).fanchor(0).lpdecay(.2).phaser(2)
  .attack(0)

$: note("{f g c d a a#}%1".sub(12))
  .gain(1).s("supersaw").lpenv(4).lpf(200).fanchor(0).lpdecay(3).phaser(3)
  .attack(0)
```


felixroos:
very cool! i guess the technique here is similar to https://github.com/tidalcycles/strudel/pull/1095 ?
haven't read the code in detail but I see you made this one work without touching the scheduler or the superdough call signature, so does this mean we could do the same with https://github.com/tidalcycles/strudel/pull/1095 ?

daslyfe:
Yes since the modulation speed is based on time and not cycle no changes to the scheduler were necessary it’s the same technique

daslyfe:
gonna go ahead and merge this so I can work on new modulations

felixroos:
> gonna go ahead and merge this so I can work on new modulations

👍

Issue #711, opened 2023-09-28
meta inputs


felixroos:
it would be handy to have a collection of meta inputs from the browser to control any parameter:

- mouseX, mouseY (https://strudel.tidalcycles.org/?8FI6F4CfCu28)
- scrollPosition
- battery status ( wink @nervousdata )
- ... what else?

felixroos:
battery level example: https://strudel.tidalcycles.org/?PZvABsrqsl_3 ( limited browser support: https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API )

felixroos:
control key:  https://strudel.tidalcycles.org/?KCWSnyIb0p--

felixroos:
on mobile: accelerometer https://developer.mozilla.org/en-US/docs/Web/API/Accelerometer

felixroos:
vowel keys: https://strudel.tidalcycles.org/?2xaH37Nj_YWh

nervousdata:
> battery level example: https://strudel.tidalcycles.org/?PZvABsrqsl_3 ( limited browser support: https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API )

Thank you for the example! I was just giving it another try in Chromium browser and generally it takes the value but when I try to pattern with e.g. `.speed("charge 1")` or multiply `.speed(charge*2)` I get this: 
[cyclist] error: Failed to set the 'value' property on 'AudioParam': The provided float value is non-finite.

felixroos:
> > battery level example: https://strudel.tidalcycles.org/?PZvABsrqsl_3 ( limited browser support: https://developer.mozilla.org/en-US/docs/Web/API/Battery_Status_API )
> 
> Thank you for the example! I was just giving it another try in Chromium browser and generally it takes the value but when I try to pattern with e.g. `.speed("charge 1")` or multiply `.speed(charge*2)` I get this: [cyclist] error: Failed to set the 'value' property on 'AudioParam': The provided float value is non-finite.

ah yes, that's because charge is already a pattern. you could do `charge.mul(2)` instead of `charge*2`.
Using it inside mini notation is not possible because strudel does not have a way to use variables in mini notation (yet),
so `"charge 1"` can only be written as `seq(charge, 1)` atm

PR #1112, opened 2024-05-25
Add the mousex and mousey signal

Enelg52:
Add two new signal sources generated by the user's mouse x and y coordinates.
```js
note("a2*8").sound("square").lpf(mousex.range(200,2000))
```

### DIFF
--- a/packages/core/signal.mjs
+++ b/packages/core/signal.mjs
@@ -86,7 +86,6 @@ export const tri2 = fastcat(isaw2, saw2);
 
 export const time = signal(id);
 
-
 /**
  *  The mouse's x or y position value ranges from 0 to 1.
  *
@@ -96,7 +95,8 @@ export const time = signal(id);
  * n(mousey.segment(4).range(0,7)).scale("C:minor")
  *
  */
-let _mouseY=0,_mouseX=0
+let _mouseY = 0,
+  _mouseX = 0;
 document.onmousemove = (e) => {
   _mouseY = e.clientY / document.body.clientHeight;
   _mouseX = e.clientX / document.body.clientWidth;
--- a/packages/core/signal.mjs
+++ b/packages/core/signal.mjs
@@ -87,25 +87,36 @@ export const tri2 = fastcat(isaw2, saw2);
 export const time = signal(id);
 
 /**
- *  The mouse's x or y position value ranges from 0 to 1.
- *
+ *  The mouse's x position value ranges from 0 to 1.
+ * @name mousex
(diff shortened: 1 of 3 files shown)


Enelg52:
Will add some more signal types as suggested here #711

packages/core/signal.mjs
```diff
@@ -86,6 +86,27 @@ export const tri2 = fastcat(isaw2, saw2);
...
+ * @return {Pattern}
+ * @example
+ * n(mousex.segment(4).range(0,7)).scale("C:minor")
+ * n(mousey.segment(4).range(0,7)).scale("C:minor")
+ *
+ */
+let _mouseY = 0,
+  _mouseX = 0;
+if (typeof window !== 'undefined') {
+  document.onmousemove = (e) => {
```
felixroos, line 101: maybe do `document.addEventListener('click', () => {})` here, because onmousemove might get overwritten?

packages/core/signal.mjs
```diff
@@ -86,6 +86,27 @@ export const tri2 = fastcat(isaw2, saw2);
...
+    _mouseY = e.clientY / document.body.clientHeight;
+    _mouseX = e.clientX / document.body.clientWidth;
+  };
+}
+
+export const mousey = signal(() => _mouseY);
```
felixroos, line 107: i never thought about using `signal` for outside references.. so far i've used the `ref` function but maybe that isn't even needed

packages/core/signal.mjs
```diff
@@ -86,6 +86,27 @@ export const tri2 = fastcat(isaw2, saw2);
...
+export const mousex = signal(() => _mouseX);
```
felixroos, line 108: maybe add `mouseX` and `mouseY` aliases?

felixroos:
> Will add some more signal types as suggested here #711

maybe in a separate PR?

packages/core/signal.mjs
Enelg52: If i do so, the value only change when i click. Shouldn't a signal constantly be changed?

felixroos: sry I mistyped, I meant `mousemove` instead of `click`

packages/core/signal.mjs
Enelg52: Done

Enelg52:
Docs should be updated under https://strudel.cc/learn/signals/

felixroos:
looks good, thank you!

PR #1116, opened 2024-05-28
Fixes drawPianoroll import in codemirror example

giohappy:
`drawPianoroll` is exported by `@strudel/draw` now.


felixroos:
thanks for fixing :) you'd need to run `pnpm i` once more and commit the `pnpm-lock.yaml` to fix the test run

giohappy:
lock file updarted

PR #1121, opened 2024-05-30
[BUG FIX] Audio worklets sometimes dont load

daslyfe:
Race condition prevented from loading sometimes before being being played leading to silence or a messed up audio context state

### DIFF
--- a/packages/superdough/superdough.mjs
+++ b/packages/superdough/superdough.mjs
@@ -27,6 +27,7 @@ export function getSound(s) {
 export const resetLoadedSounds = () => soundMap.set({});
 
 let audioContext;
+export const isAudioInitialized = () => audioContext != null;
 
 export const setDefaultAudioContext = () => {
   audioContext = new AudioContext();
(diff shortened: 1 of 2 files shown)


felixroos:
it might be simpler to promise cache `initAudioOnFirstClick` and do `await initAudioOnFirstClick` inside `beforeEval`? I think the presence of audioContext doesn't mean that the worklets are loaded

felixroos:
so the problem was the first event getting played before audioworklets being ready?

daslyfe:
> so the problem was the first event getting played before audioworklets being ready?

yes it was two things

1. there was no awaiting in the loadWorklets function, so it was sometimes not initialized before it was referenced
2. the initAudioOnFirstClick happens at the same time as you hit the play button if you have not clicked anything else yet

daslyfe:
> it might be simpler to promise cache `initAudioOnFirstClick` and do `await initAudioOnFirstClick` inside `beforeEval`? I think the presence of audioContext doesn't mean that the worklets are loaded

the audioContext being nonnull implies that the worklets have been loaded, because initializing the audioContext now awaits for the worklets to finish loading

PR #1125, opened 2024-06-02
doc: visual functions + refactor onPaint

felixroos:
document visual functions along with some fixes to allow using them in the mini repl


felixroos:
should be ready now..

PR #1130, opened 2024-06-12
Phase vocode

daslyfe:
s("bd!2 sd bd").stretch("<0 .25 -1>")


packages/superdough/fft.js
felixroos, line 1: have you thought about adding this as a dependency? or is this only a smaller part of https://www.npmjs.com/package/fft.js  ?

daslyfe: its just a small part of the whole package. I also think that it is super useful to be able to modify and expand it in the future.

PR #1132, opened 2024-06-14
Udels (MultiFrame Strudel) Revisited

daslyfe:
To try in dev mode visit: 
http://localhost:4321/udels#cygiYmQhNCIpCg%3D%3D,c3RhY2soCiAgcygiaGghMTYiKQop

- base app unaffected, new page has been added at /udels with multiwindow functionality
- Code stored in URL strings passed from child to parent window
- each instance can have its own custom settings and saved patterns
- I broke up some parts of the Repl into composable components so the editor jsx code was cleaner/more transposable.

image


packages/core/evaluate.mjs
```diff
@@ -39,6 +39,10 @@ function safeEval(str, options = {}) {
 
 export const evaluate = async (code, transpiler, transpilerOptions) => {
   let meta = {};
+  //post to iframe parent (like Udels) if it exists...
```
felixroos, line 42: could this also be done inside Repl.jsx afterEval ? seems a bit too core to do it here + the handling of the url hash is also there for the normal repl

website/src/components/Udels/UdelsEditor.jsx
```diff
@@ -0,0 +1,34 @@
...
+//  editorRef:  React.MutableRefObject<HTMLElement | null>,
+//  error: Error
+//  init: () => void
+// }
+
+export default function UdelsEditor(Props) {
+  const { context, containerRef, editorRef, error, init } = Props;
+  const { pending, started, handleTogglePlay } = context;
+  return (
+    <ReplContext.Provider value={context}>
```
felixroos, line 21: maybe it would be simpler to pull out this Provider to Repl.jsx and throw in the stuff needed? Then useContext could be used anywhere without "prop drilling". I tend to get confused when props are passed more than 2 levels deep, but I won't say I haven't done it myself :P

website/src/settings.mjs
```diff
@@ -29,7 +30,15 @@ export const defaultSettings = {
...
-export const settingsMap = persistentMap('strudel-settings', defaultSettings);
+let search = null;
+if (typeof window !== 'undefined') {
+  search = new URLSearchParams(window.location.search);
+}
+// if running multiple instance in one window, it will use the settings for that instance. else default to normal
+const instance = parseInt(search?.get('instance') ?? '0');
+const settings_key = `strudel-settings${instance > 0 ? instance : ''}`;
+
+export const settingsMap = persistentMap(settings_key, defaultSettings);
```
felixroos, line 41: could be cool the clone the settings of the main repl or latest udel repl instead of using defaultSettings? tho it's only relevant on first use

felixroos:
just tested it and it's great! feels like dj mode :)

website/src/settings.mjs
daslyfe: Yes that makes sense

packages/core/evaluate.mjs
daslyfe: sure I can give that a try

PR #1139, opened 2024-07-03
New tools for long-form compositions: setLoop() and cyclecounter()

eefano:
**setLoop(start, length)** 
alters the core behaviour of Cyclist (like setCps does): playback is always started at the cycle number 'start' (default=0). When a 'length' number of cycles have elapsed, the cycles counter is reset to 'start'. The parameters are floored (decimal part of the number is ignored). When 'length' is zero (default) the playback is not looped.
Evaluating setLoop() without parameters restores the original behaviour.

**cyclecounter()**
this widget displays the integer part of 'time' value (that is, the current cycle number) in the bottom-right part of the screen. Optional parameters allow the customization of font and color. The 'div' option divides the counter value by a constant. For example: when all measures are 4/4 and 1beat=1cycle, it can be useful to count the measure number by passing {div: 4}

The actual implementation is not very precise at the edges and the visual feedback is janky (I'm hoping for your help in this regard), but it's still very useful for fast-iterating a middle part of a long composition, without altering the overall structure every time.

Example, this plays C D E in loop:
```
setCps(120/60)
setLoop(2,3)
note("<a4 b4 c4 d4 e4 f4 g4>").cyclecounter()
```

Ode to Joy, but only the 3rd and 4th measure (the counter displays Cycles/4):
```
setCps(120/60)
setLoop(2*4, 2*4)
note("<e4!2 f4 g4!2 f4 e4 d4 c4!2 d4 e4 [e4@3 d4]@2 d4@2>").cyclecounter({div:4})
```

Let me know what do you think ... thank you as always!

### DIFF
--- a/packages/codemirror/widget.mjs
+++ b/packages/codemirror/widget.mjs
@@ -133,3 +133,8 @@ registerWidget('_pitchwheel', (id, options = {}, pat) => {
   const ctx = getCanvasWidget(id, options).getContext('2d');
   return pat.pitchwheel({ ...options, ctx, id });
 });
+
+registerWidget('_cyclecounter', (id, options = {}, pat) => {
+  const ctx = getCanvasWidget(id, options).getContext('2d');
+  return pat.cyclecounter({ ...options, ctx, id });
+});
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -23,6 +23,8 @@ export class Cyclist {
     this.beforeStart = beforeStart;
     this.cps = 0.5;
     this.num_ticks_since_cps_change = 0;
+    this.loopStart = 0;
+    this.loopLength = 0;
     this.lastTick = 0; // absolute time when last tick (clock callback) happened
     this.lastBegin = 0; // query begin of last tick
     this.lastEnd = 0; // query end of last tick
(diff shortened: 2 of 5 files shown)


felixroos:
As said earlier in discord, `ribbon` can be used to get the behavior of `setLoop`:

```js
all(x=>x.ribbon(2,3))
```

I still like the idea of having a helper function like `setLoop` that is simpler to write, but I'd rather not implement it in the cyclist to avoid too much complexity.

The problem rn is that `all` can only be used once, so if we add helper functions that uses `all` under the hood, it might happen that it gets overridden, e.g.:

```js
all(x=>x.hpf(1000))
setLoop(2,3) // <-- overwrites line above
```

To fix that, we could implement `all` in a way that it adds a transformation each time it is called. rn it looks like this:

```js
let allTransforms = [];
const all = function (transform) {
    allTransforms.push(transform)
    return silence;
  };
// later:
if (allTransforms.length) {
  for(let i in allTransforms) {
    pattern = allTransforms[i](pattern);
  } 
}
```

That would allow defining `setLoop` as

```js
let setLoop = (a,b) => all(x=>x.ribbon(a,b))
```

I like the idea of `cyclecounter`, but maybe it could be part of the DOM instead? Also seems like this should be 2 PRs.
What do you think @eefano ?

PR #1141, opened 2024-07-13
Real Legato

daslyfe:
Legato now changes the note duration relative to the gap between notes.

Screenshot 2024-07-13 at 1 05 44 PM

`s("sawtooth").euclid(11,16).legato("<1 .5>")._pianoroll()`

second array parameter is "lookahead" which is the number of cycles in the future to query for the next event
 
`s("supersaw").euclid(7,16).legato("1:<1` .15>")._pianoroll()`


daslyfe:
deleted this as the pr got messed up

PR #1142, opened 2024-07-13
Real legato

daslyfe:
Legato now changes the note duration relative to the gap between notes.

Screenshot 2024-07-13 at 1 05 44 PM

`s("sawtooth").euclid(11,16).legato("<1 .5>")._pianoroll()`

second array parameter is "lookahead" which is the number of cycles in the future to query for the next event
 
`s("supersaw").euclid(7,16).legato("1:<1` .15>")._pianoroll()`


geikha:
I've been thinking about this and I'm worried about the name collision with the original Tidal legato.

How about `articulate` ? 

https://en.wikipedia.org/wiki/Articulation_(music)
https://www.image-line.com/fl-studio-learning/fl-studio-online-manual/html/pianoroll_articulate.htm

felixroos:
> I've been thinking about this and I'm worried about the name collision with the original Tidal legato.
> 
> How about `articulate` ?
> 
> https://en.wikipedia.org/wiki/Articulation_(music) https://www.image-line.com/fl-studio-learning/fl-studio-online-manual/html/pianoroll_articulate.htm

thinking the same. articulate seems a bit abstract to me.. how about fill?

daslyfe:
> > I've been thinking about this and I'm worried about the name collision with the original Tidal legato.
> > How about `articulate` ?
> > https://en.wikipedia.org/wiki/Articulation_(music) https://www.image-line.com/fl-studio-learning/fl-studio-online-manual/html/pianoroll_articulate.htm
> 
> thinking the same. articulate seems a bit abstract to me.. how about fill?

I like fill I think that makes sense and is easy to type :) technically it is legato though and a fill means something different musically. Maybe that’s okay

felixroos:
> > > I've been thinking about this and I'm worried about the name collision with the original Tidal legato.
> > > How about `articulate` ?
> > > https://en.wikipedia.org/wiki/Articulation_(music) https://www.image-line.com/fl-studio-learning/fl-studio-online-manual/html/pianoroll_articulate.htm
> > 
> > 
> > thinking the same. articulate seems a bit abstract to me.. how about fill?
> 
> I like fill I think that makes sense and is easy to type :) technically it is legato though and a fill means something different musically. Maybe that’s okay

another alternative could be flood

geikha:
> technically it is legato though and a fill means something different musically.

well yeah but in this context it wouldn't mean "a fill" more of like "fill-in the silences"

geikha:
btw this has a problem, i've run into this in tidal and yeah it replicates in strudel too. the malformed haps list is troublesome when combined with other complex hap-changing functions such as rev:

```js
$: s("supersaw")
  // .euclid(11,16)
  .struct("1 0 0 1 1").fast(2)
  // .beat("0:1:3:4:6:5", 8).fast(2)
  .legato("1").rev()
  ._pianoroll()
 ```

**Comment by daslyfe** [comment]
> btw this has a problem, i've run into this in tidal and yeah it replicates in strudel too. the malformed haps list is troublesome when combined with other complex hap-changing functions such as rev:
> 
> ```js
> $: s("supersaw")
>   // .euclid(11,16)
>   .struct("1 0 0 1 1").fast(2)
>   // .beat("0:1:3:4:6:5", 8).fast(2)
>   .legato("1").rev()
>   ._pianoroll()
> ```

hmm yeah im not sure what the solve is for this edge case, it still works if you put rev before legatio at least.. `$: s("supersaw")
  // .euclid(11,16)
  .struct("1 0 0 1 1").fast(2).rev()
  // .beat("0:1:3:4:6:5", 8).fast(2)
  .legato("1")
  ._pianoroll()`

**Comment by daslyfe** [comment]
> > I've been thinking about this and I'm worried about the name collision with the original Tidal legato.
> > How about `articulate` ?
> > https://en.wikipedia.org/wiki/Articulation_(music) https://www.image-line.com/fl-studio-learning/fl-studio-online-manual/html/pianoroll_articulate.htm
> 
> thinking the same. articulate seems a bit abstract to me.. how about fill?

looks like fill is a reserved param for the pianoroll, does it need to be?

**Comment by geikha** [comment]
@daslyfe 

> hmm yeah im not sure what the solve is for this edge case

here's my solution from the haskell side: a function which, given any event (hap), it encloses it into the original query arc (span), if it shouldn't be there it'll return nothing. in the js side returning null will probably be good enough

```
encloseEvent :: Ord a => ArcF a -> EventF (ArcF a) b -> Maybe (EventF (ArcF a) b)
encloseEvent _ (Event _ Nothing _ _) = Nothing
encloseEvent (Arc as ae) ev@(Event ctx (Just (Arc ws we)) part val)
  | ws >= ae = Nothing
  | ws >= as && we <= ae = Just ev
  | ws >= as && we >  ae = Just ev { part = Arc ws ae }
  | ws <  as && we <= ae = Just ev { part = Arc as we }
  | ws <  as && we >  ae = Just ev { part = Arc as ae }
  | otherwise = Nothing
```

**Comment by geikha** [comment]
ok here's the issue with the lookahead scenario: there are times where some haps will be lots only because their original times weren't on the query. what i did for quantization for tidal doesn't run into this scenario, since what i do there is, for every query span, query the whole cycle (floor and ceiling of the current query span, basically), and then enclose it. if you make sure you always query from the cycle start, you're good to go to then enclose it and have it well-formed

**Comment by geikha** [comment]
Other name idea `gapless` https://github.com/tidalcycles/Tidal/issues/761

**Comment by felixroos** [comment]
what are we doing with this one?

**Comment by daslyfe** [comment]
> what are we doing with this one?

I think it is a super useful function, I just need help solving the edge cases and doing it in a more "correct" way

**Comment by yaxu** [comment]
I started having a go at this, but it seems the same approach as yours @daslyfe, and it's also broken by `rev()`

```
// "real legato"

Pattern.prototype.sortHapsByWhole = function () {
    return this.withHaps((haps) =>
      haps.sort((a, b) =>
        a.whole.begin.sub(b.whole.begin)
        .or(a.whole.end.sub(b.whole.end))
        .or(a.part.begin.sub(b.part.begin))
        .or(a.part.end.sub(b.part.end))
      ),
    );
  }

const rl = register('rl', function (pat) {
  return new Pattern(function(state) {
    const lookahead = 1;
    pat = pat.sortHapsByWhole();
    const haps = pat.query(state.withSpan(span => span.withEnd(end => end.add(lookahead))));
    // TODO - make elements unique
    const onsets = haps.map(hap => hap.whole.begin);
    const newHaps = [];
    for (const hap of haps) {
      if (hap.part.begin.lt(state.span.end)) {
        const next = onsets.find(onset => onset.gte(hap.whole.end));
        if (next) {
          const whole = new TimeSpan(hap.whole.begin, next);
          const part = new TimeSpan(hap.part.begin, next.gt(state.span.end) ? state.span.end : next);
          newHaps.push(new Hap(whole, part, hap.value, hap.context, hap.stateful));
        }
        else {
          newHaps.push(hap);
        }
      }
    }
    return newHaps;
  });
});

note("c(3,8) e - f").rl().slow(2).sound("folkharp")
  .pianoroll()
```

**Comment by yaxu** [comment]
Maybe the bug is in rev... https://discord.com/channels/779427371270275082/797978366023434240/1303280618863329280

**Comment by yaxu** [comment]
Ah no it just needs to do look behind as well as look ahead, otherwise hap tails are lost

**Comment by yaxu** [comment]
Ok this proof-of-concept works! It is now querying two cycles worth of events, which I'd expect would eat a lot of cpu in some cases but I haven't being able to find any immediately playing with high `chop` values..

I also didn't implement duration factors yet. I guess that should be clamped at 1, otherwise the query window would need to be opened wider.

```js
// "real legato"

const rl = register('rl', function (pat) {
  return new Pattern(function(state) {
    const lookbothways = 1;
    // Expand the query window
    const haps = pat.query(state.withSpan(span => new TimeSpan(span.begin.sub(lookbothways), span.end.add(lookbothways))));
    const onsets = haps.map(hap => hap.whole.begin)
      // sort fractions
      .sort((a, b) => a.compare(b))
      // make unique
      .filter((x, i, arr) => i == (arr.length-1) || x.ne(arr[i+1]));
    const newHaps = [];
    for (const hap of haps) {
      // Ingore if the part starts after the original query
      if (hap.part.begin.gte(state.span.end)) {
        continue;
      }

      // Find the next onset, to use as an offset
      const next = onsets.find(onset => onset.gte(hap.whole.end));
      
      // Ignore if the part ended before the original query, and hasn't expanded inside
      if (hap.part.end.lt(state.span.begin) && next.lte(state.span.begin)) {
        continue;
      }
      
      const whole = new TimeSpan(hap.whole.begin, next);
      // Constrain part to original query
      const part = new TimeSpan(hap.part.begin.max(state.span.begin), next.min(state.span.end));
      newHaps.push(new Hap(whole, part, hap.value, hap.context, hap.stateful));
    }
    return newHaps;
  });
});

note("c(3,8) e - f").rl().sound("folkharp").palindrome()
  .pianoroll()
```

yaxu:
It's pretty fun!

https://strudel.cc/?-nuZXA5YmTey

There's still a bug somewhere though I think, gaps appear sometimes..
https://strudel.cc/?9k-DxB18XpbC

yaxu:
@daslyfe Do you fancy picking this up or shall I try and finish it ? I'm travelling the next days

daslyfe:
> @daslyfe Do you fancy picking this up or shall I try and finish it ? I'm travelling the next days

I have 4 shows Im preparing for this week, but Ill be able to check this out next Monday! Thank you for helping to move this forward

yaxu:
Oh nice, good luck!

PR #1160, opened 2024-08-11
Create audio target selector for OSC/Superdirt

daslyfe:
Allows you to set the default target to osc so you do not have to write .osc() after everything if you are using superdirt as an audio backend. (webaudio is the default of course) Also: 
- fixes a bug in webkit browsers where confirm would happen automatically regardless of user choice 
- fixes OSC jitter for the web version of strudel
Screenshot 2024-08-11 at 12 40 27 PM


yaxu:
This looks good! I guess someone might select this by mistake, and then puzzle over why strudel stops working.. But they could always click the 'restore default settings' button in that case.

daslyfe:
> This looks good! I guess someone might select this by mistake, and then puzzle over why strudel stops working.. But they could always click the 'restore default settings' button in that case.

I added 
Screenshot 2024-08-31 at 11 35 18 PM
a warning message that points to the docs, and updated the docs as well :)

daslyfe:
I plan on merging this tomorrow morning after extensive testing

PR #1166, opened 2024-08-18
Add global Audio Volume setting to REPL

netux:
Allows for adjusting the global volume/gain of the REPL. Set default global value to 50%.

- Acts similarily to `gain()`, but without modifying the code.
- The volume slider uses a logarithmic scale, so it adjusts better to human sound perception.
- Volume defaults to 50%, which was ideal for me. It can be raised if needed.

I also wanted to add a volume slider to the MiniRepl. But because multiple MiniRepl's can be on a page at the same time, while this setting is global, really makes it hard to implement.

### DIFF
--- a/packages/superdough/superdough.mjs
+++ b/packages/superdough/superdough.mjs
@@ -143,6 +143,14 @@ export function initializeAudioOutput() {
   destinationGain.connect(audioContext.destination);
 }
 
+export function setGlobalGain(gain) {
+  if (destinationGain == null) {
+    initializeAudioOutput();
+  }
+
+  destinationGain.gain.value = gain;
+}
+
 // input: AudioNode, channels: ?Array<int>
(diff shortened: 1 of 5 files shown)


yaxu:
Awesome, I've been finding 50% works well for me as well..

yaxu:
The recently merged PR #1160 added an 'audio engine target' option. The audio volume slider doesn't have any effect when the 'osc' option is selected. Would it be possible to hide or grey out the audio volume slider when webaudio isn't selected? I don't think it would make too much sense to have the slider change the gain control or something like that, as it would need a different default to match superdirt (probably 1).

netux:
Sure. How about that?
audio engine target set to osc, audio volume slider disabled

netux:
Got a few suggestions from Discord (namely, the new algorithm to calculate resulting gain), as well as a couple things I found while retesting.

PR #1181, opened 2024-09-06
create macro params for oscillators

daslyfe:
The intention is to make generic macro params z1-z4 for synth sounds. this concept has been used very successfully in mutable instruments modules as well as electron boxes.

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -74,6 +74,49 @@ export function registerControl(names, ...aliases) {
 export const { s, sound } = registerControl(['s', 'n', 'gain'], 'sound');
 
 /**
+ * generic macro param for a sound synth
+ *
+ * @name z1
+ * @synonyms z
+ * @param {number | Pattern}
+ * @example
+ * n(run(8)).scale("D:pentatonic").s("supersaw").x(".01 .75").release(0.5)
+ *
+ */
+export const { z1, z } = registerControl(['z1', 'z2', 'z3', 'z4'], 'z');
+
+/**
+ * generic macro param for a sound synth
+ *
+ * @name z2
+ * @param {number | Pattern}
+ * @example
+ * n(run(8)).scale("D:pentatonic").s("supersaw").x2("2 .7").release(0.5)
+ *
+ */
+export const { z2 } = registerControl('z2');
+
+/**
+ * generic macro param for a sound synth
+ *
+ * @name z3
+ * @param {number | Pattern}
+ * @example
+ * n(run(8)).scale("D:pentatonic").s("supersaw").x3(".1 .75").release(0.5)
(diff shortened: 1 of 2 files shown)


packages/core/controls.mjs
```diff
@@ -73,6 +73,49 @@ export function registerControl(names, ...aliases) {
...
+export const { z4 } = registerControl('z3');
+
+/**
+ * generic macro param for a sound synth
+ *
+ * @name x4
+ * @param {number | Pattern}
+ *
+ */
+export const { x4 } = registerControl('z4');
```
felixroos, line 117: typo here :)

felixroos:
do i understand it correctly, that each z param has a named counterpart, and it is mostly a shortcut + a way to see what happens? I generally like the idea, but I wonder how this should be documented best, especially if some sounds have it and some don't

felixroos:
also, i think we need to be careful that the range doesn't differ much between z controls, too prevent something like

1. you start with sound A and z1(400) 
2. you switch to sound B, your ears will fall off because z1 400 for that sound is massively out of the pleasant range

not sure how to do this, maybe let 0-1 be the safe range? would that mean we have to rescale z params to their named counterpart?  just some thought

daslyfe:
> also, i think we need to be careful that the range doesn't differ much between z controls, too prevent something like
> 
>     1. you start with sound A and z1(400)
> 
>     2. you switch to sound B, your ears will fall off because z1 400 for that sound is massively out of the pleasant range
> 
> 
> not sure how to do this, maybe let 0-1 be the safe range? would that mean we have to rescale z params to their named counterpart? just some thought

I think normalizing the z parameters between 0 and 1 is a great idea. I updated the pr. Also, I think sometimes they will refer to existing params where it makes sense (like with supersaw). However, there might be times when having a specific control for an oscillator might not make sense in which case you would only use z to control it.

Bubobubobubobubo:
> you switch to sound B, your ears will fall off because z1 400 for that sound is massively out of the pleasant range

The beauty of Elektron boxes or Mutable is that this never happens. Whatever mode you switch into, the same set of parameters will still produce something that you can work with, albeit different. Of course, the difference can be drastic if you jump from a tonal VS atonal sound source, but given that you have an envelope and filter, it will still be usable.

> I wonder how this should be documented best

Some of the models used in macro-oscillators are kind of "self-documented". You can imagine having a model called "3osc" whose macros will help you jump between different frequency tunings around a central frequency (z1), handle the oscillator detune (z2), interpolate between waveforms (z3) and whatever (z4). It will still be 3-oscillator voice nonetheless.

The downside of this approach is that you need a lot of carefully written and optimized DSP to generate a pleasant collection. In this regard, the Mutable code itself could help.

Issue #1186, opened 2024-09-18
cat method shows up twice with two different explanations in the API reference.


kdiab:
`cat` shows up twice with two different explanations in the reference.
![image](https://github.com/user-attachments/assets/c20faea4-ec6a-4a68-a4d5-68ab26d7663d)

https://github.com/tidalcycles/strudel/blob/main/packages/core/pattern.mjs#L844
https://github.com/tidalcycles/strudel/blob/main/packages/core/pattern.mjs#L1402

Would it be better if those were combined?

PR #1189, opened 2024-09-21
remove redundant example for cat, update snapshot

kdiab:
This PR is for issue  #1186

`cat` was documented twice causing it to appear twice in the API reference. This change removes the less descriptive example and keeps the other one. 

image

### DIFF
--- a/packages/core/pattern.mjs
+++ b/packages/core/pattern.mjs
@@ -809,15 +809,6 @@ export class Pattern {
   //////////////////////////////////////////////////////////////////////
   // Multi-pattern functions
 
-  /**
-   * Stacks the given pattern(s) to the current pattern.
-   * @name stack
-   * @memberof Pattern
-   * @example
-   * s("hh*4").stack(
-   *   note("c4(5,8)")
-   * )
-   */
   stack(...pats) {
     return stack(this, ...pats);
   }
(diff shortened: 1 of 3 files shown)


felixroos:
thanks for your contribution. It looks like this example you've removed is still referenced (as Pattern.cat) in https://strudel.cc/learn/factories/#cat , so maybe add it there as a MiniRepl instead?

kdiab:
Yeah I think that's a good idea! This looks more informative.

image

kdiab:
Found the same bug for `seq` and `stack`, going to combine with this PR. 
image
image

felixroos:
>  It looks like this example you've removed is still referenced (as Pattern.cat) in https://strudel.cc/learn/factories/#cat 

have you looked at this?  seq and stack might have refs as well

kdiab:
Yes! That's where I got the examples from 😊

felixroos:
i mean have you checked if the factories page still looks as before? this: https://github.com/tidalcycles/strudel/blob/e409086b34c9a226e518a190ed86cc2510c640be/website/src/pages/learn/factories.mdx?plain=1#L38 just to double check

kdiab:
Removed unnecessary instances of functions and fixed the factories folder.

image
image
image
image
image

felixroos:
thanks, this now looks good!

PR #1192, opened 2024-10-04
Mini mininotation parser

yaxu:
Backported from https://github.com/felixroos/idlecycles


felixroos:
this overlaps with https://github.com/tidalcycles/strudel/pull/1311 . even without mondo, i think it would still be desirable to have a mini parser that is more self contained / easier to change and adapt

PR #1209, opened 2024-10-30
Apply `all` function to individual patterns rather than final stack

yaxu:
This fixes `all` to work with 'pulsewise' functions.


felixroos:
I've just noticed this doesn't play well with pianoroll: https://strudel.cc/#JDogbm90ZSgiYSIpCgokOiBub3RlKCJlIikKCmFsbCh4PT54LnBpYW5vcm9sbCh7bGFiZWxzOjEsZm9sZDowfSkp

would be good to have 2 functions for both behaviours i think

PR #1229, opened 2024-12-23
Make `all()` post-stack again, and add `each()` for pre-stack

yaxu:
Since #1209 changed `all()` to be applied to patterns separately, the old 'post-stack' behaviour has been missed, for things like applying a single `.pianoroll()` to everything. This brings the old behaviour back as `.together()`.

### DIFF
--- a/packages/core/repl.mjs
+++ b/packages/core/repl.mjs
@@ -168,7 +168,7 @@ export function repl({
         pattern = allTransform(pattern);
       }
       if (togetherTransform) {
-        pattern = togetherTransform(pattern)
+        pattern = togetherTransform(pattern);
       }
       if (!isPattern(pattern)) {
         const message = `got "${typeof evaluated}" instead of pattern`;
--- a/packages/core/repl.mjs
+++ b/packages/core/repl.mjs
@@ -63,11 +63,13 @@ export function repl({
   let pPatterns = {};
   let anonymousIndex = 0;
   let allTransform;
+  let togetherTransform;
 
   const hush = function () {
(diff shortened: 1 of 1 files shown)


felixroos:
could we maybe swap the meaning, so all stays the stack thing? there are a bunch of patterns out that use all(x=>x.pianoroll()) or all(x=>x.scope()) which breaks when calling the function multiple times

yaxu:
I don't think all vs together would make sense the other way around. We could have `each` for individual and `all` for stacked though.

yaxu:
(In any case we should support toplevel functions for visualisations so `x(pianoroll)` works.)

yaxu:
Went ahead with original flavour `all` and new `each` and making that live, as the new `all` was a bit broken anyway.

A `pianoroll` function is already exported - am I right though that this is mostly an internal function and could be renamed to make room for `pianoroll = x => x.pianoroll()`? 

https://github.com/tidalcycles/strudel/blob/9bf0f624f5ec3d36f03a0fb18745bd3809ca3dab/packages/draw/pianoroll.mjs#L104-L268

felixroos:
> Went ahead with original flavour `all` and new `each` and making that live, as the new `all` was a bit broken anyway.

thanks!

> A `pianoroll` function is already exported - am I right though that this is mostly an internal function and could be renamed to make room for `pianoroll = x => x.pianoroll()`?

yes it's more internal i'd say. maybe call it drawPianoroll or something. if we add this, the same could with the other draw functions for consistency, tho that can also wait

PR #1230, opened 2024-12-24
understand voicings page

felixroos:


daslyfe (approved):
Thank you for doing this!! It will be much easier to teach now

TodePond (approved):
very helpful and interesting! 

i followed through and noted down and questioned I had along the way. 

- I didn't understand what you meant by "the voicings will be automatically generated with smooth voice leading" but then you got to it later on and it was fine. maybe adding a "(more on this later)" or something would have helped me to keep going. 
- when you say "- ^ is the same as M, for example C^7 = CM7", I assume that C7 is also the same? or am I wrong? 

do whatever you want with this user test, thanks for the write up :)

felixroos:
thanks for the feedback!

I've tweaked a few sentences to hopefully make it clearer. 

> when you say "- ^ is the same as M, for example C^7 = CM7", I assume that C7 is also the same? or am I wrong?

this is indeed a bit confusing, but unfortunately the "norm": C^ = CM = C . but C^7 !== C7 ... C^7 = C E G B but C7 = C E G Bb or generally "^7" = 0 4 7 11 and "7" = 0 4 7 10 . It might make sense to add a little bit more theoretical background about where these chord symbols come from / what they mean.

bwagner:
Thank you, @felixroos, for your [answer](https://discord.com/channels/779427371270275082/1191011773311107153/1321832214446477364) on discord about how to implement Drop2-voicings!


Maybe it helps to understand if the voicing dictionary in the example would be broken down:

```
addVoicings('house', {
'': ['7 12 16', '0 7 16', '4 7 12'],
'm': ['0 3 7']
})
```

The first key is the default (empty string), hence the major triad.
The provided voicings are:
7 = 7 half steps = pure 5th
12 = 12 half steps = pure octave
16 = 16 half steps = octave above major third

I guess the order of these offsets is irrelevant to the result since they'll sound simultaneously. Still, this might be worth mentioning.

PR #1240, opened 2025-01-12
Try to solve method composition with proxy objects

yaxu:


TodePond:
i don't have enough context on the strudel innards to comment generally, but I'll just throw out a heads-up that I've found `Proxy` to be pretty slow, so might be worth testing out any performance impacts

yaxu:
> i don't have enough context on the strudel innards to comment generally, but I'll just throw out a heads-up that I've found `Proxy` to be pretty slow, so might be worth testing out any performance impacts

Thanks for the heads-up, I tried adding a benchmark, seems OK, but maybe it causes problems in some cases..

image

yaxu:
The latest commit supports `set.n(3)` instead of `set(n(3))` but not yet `set.out.n(3)` instead of `set.out(n(3))`.

felixroos:
sorry that i haven't really given you feedback on this so far, also not very sure about it. it feels a bit crazy to use a proxy for every method. from my current viewpoint of uzulanging i'd say we shouldn't make the metaprogramming too crazy, as we can get a more ideal DSL when going full uzu. in mondo, the example snippet would look like this:

```plaintext
sound [bd md cp] .every 3 (.fast 2 .n [3|2] .chop 4 .jux rev)
```

these problems are much easier to solve when having more control over the language level.
but of course the js dsl isn't going anywhere anytime soon, so it's still desirable to improve it, i'm just a bit worried that Proxies might get us into trouble later (although we could revert it if that's the case). i think i remember using a library (mobx) that uses proxies and they made it very hard to debug them in the dev tools, but Patterns are not really debuggable in the dev tools anyway so that might not matter much. still i think Proxies are a rather magical feature of js and i don't fully understand them enough to know the consequences, but we could also take the risk and see what happens

PR #1245, opened 2025-01-18
Add bank aliasing and case insensitivity

TodePond:
## What is this pull request?

This is a spicy pull request that adds an `aliasBank` function. It lets you create an alias for any bank of sounds.

image

This pull request **also** adds lots of aliases by defaults. For example, "TR909" is an alias for "RolandTR909". Check out the aliases JSON file for all the aliases.

image

This pull request **also** removes cases-sensitivity from ALL SOUNDS. This is a big breaking change.
If this isn't wanted at this point, let's remove it and discuss within a future pull request.

image

## Release notes

- Added the `aliasBank` function that lets you create aliases for banks.
- Added lots of default aliases for drum machine banks, eg: "TR909" for "RolandTR909".
- Sounds are now case insensitive.

## Future work

- Make the sounds interface nicer and easier.
- Move the JSON hosted elsewhere into the strudel repo (see comments).


.prettierignore
```diff
@@ -11,3 +11,4 @@ pnpm-lock.yaml
 pnpm-workspace.yaml
 **/dev-dist
 website/.astro
+!tidal-drum-machines.json
```
TodePond, line 14: also enabled formatting for this file

yaxu:
How does this look in the sound browsing interface?

I find the sound bank implementation difficult to teach in workshops, especially where 'sounds' are already a sample set and 'banks' are a level above but don't appear that way in the sound browser. Improving this is probably out of scope for this PR but we should try not to make things more confusing.

Agreed the shorter names are better and would like to deprecate the long ones. Perhaps we could also go for all lowercase with the short versions.

felixroos:
> How does this look in the sound browsing interface?
> 
> I find the sound bank implementation difficult to teach in workshops, especially where 'sounds' are already a sample set (via `n`) and 'banks' are a level above but don't appear that way in the sound browser. Improving this is probably out of scope for this PR but we should try not to make things more confusing.
> 
> Agreed the shorter names are better and would like to deprecate the long ones. Perhaps we could also go for all lowercase with the short versions.

yes the sound tab needs renovation! 

another hot take hot take hot take: we could make use of the fact that all drum samples get new names and give a little tlc to the samples (e.g. 808 defaults are weird) + maybe mix them differently (hihats quieter etc..) ? if we do it now, before people use them, we won't get a breaking change..

yaxu:
> another hot take hot take hot take: we could make use of the fact that all drum samples get new names and give a little tlc to the samples (e.g. 808 defaults are weird) + maybe mix them differently (hihats quieter etc..) ? if we do it now, before people use them, we won't get a breaking change..

Yes that's a nice idea! Then every time we have to type RolandTR808 we're more motivated to do some sample mastering..

yaxu:
Strictly speaking Visco probably isn't a manufacturer name because Visco Space Drum never existed https://gearspace.com/board/electronic-music-instruments-and-electronic-music-production/390269-visco-space-drum.html

TodePond:
> How does this look in the sound browsing interface?

good point, i hadn't thought of that. thanks for pointing it out. 

this pr could probably filter out one set of names in the menu, and a future pull request could improve the overall ux of the drum sound menu

> Perhaps we could also go for all lowercase with the short versions.

even better could be to make it case insensitive?

> Visco Space Drum probably never existed

woah!

I'm not wedded to any choices here, will happily merge or make any changes no questions asked

---

just a heads up to both of you that the shortened list misses off a couple that i couldn't figure out how to shorten

felixroos:
an alternative take on this would be to handle aliases after the json has been processed. here's an example [aliasBank function](https://strudel.cc/#bGV0IGFsaWFzQmFuayA9IChhbGlhcywgbmFtZSkgPT4gewogIGNvbnN0IGVudHJpZXMgPSBPYmplY3QuZW50cmllcyhzb3VuZE1hcC5nZXQoKSkgCiAgY29uc3Qgc291bmRzID0gZW50cmllcwogIC5maWx0ZXIoKFtzb3VuZF0pID0%2BIHNvdW5kLnN0YXJ0c1dpdGgobmFtZSsnXycpKQogIC5mb3JFYWNoKChbc291bmQsdl0pID0%2BIHsKICAgIGNvbnN0IFtfLHN1ZmZpeF0gPSBzb3VuZC5zcGxpdCgnXycpCiAgICBlbnRyaWVzLnB1c2goW2FsaWFzKydfJytzdWZmaXgsIHZdKTsKICB9KQogIGNvbnN0IG5leHQgPSBPYmplY3QuZnJvbUVudHJpZXMoZW50cmllcykKICBzb3VuZE1hcC5zZXQobmV4dCk7Cn0KCgphbGlhc0JhbmsoJ1RSOTA5JywnUm9sYW5kVFI5MDknKQoKJDogcygiYmQgaGgiKS5iYW5rKCdUUjkwOScpCiQ6IHMoIi0gY3AiKS5iYW5rKCdSb2xhbmRUUjkwOScp) . this would save file size + it's usable with from disk autogenerated strudel.json files á la [@strudel/sampler](https://www.npmjs.com/package/@strudel/sampler) (based on discussion in strudel-innards)

felixroos:
@TodePond do you want to add the aliasBank function above? would be nice to get this into the next version

TodePond:
sure thing! do you mean just the function? (for people to use) 

or also all the aliases? (called with the function) 

hope that makes sense

felixroos:
> sure thing! do you mean just the function? (for people to use)
> 
> or also all the aliases? (called with the function)
> 
> hope that makes sense

i'd say also the aliases called with the function. maybe make them lowercase as well? there might also be a way to improve / simplify it for setting up multiple aliases at once but not super important

felixroos:
> yes!
> 
> when are you planning / hoping to do the release by the way?

soonish.. when the next wave of PRs is merged, but not sure when that will be exactly..

packages/repl/prebake.mjs
```diff
@@ -21,6 +21,9 @@ export async function prebake() {
   );
   // load samples
   const ds = 'https://raw.githubusercontent.com/felixroos/dough-samples/main/';
+
+  // TODO: move this onto the strudel repo
+  const ts = 'https://raw.githubusercontent.com/todepond/samples/main/';
```
TodePond, line 26: https://github.com/TodePond/samples/blob/main/tidal-drum-machines-alias.json

packages/repl/prebake.mjs
```diff
@@ -36,6 +39,8 @@ export async function prebake() {
     samples(`${ds}/EmuSP12.json`),
     samples(`${ds}/vcsl.json`),
   ]);
+
+  aliasBank(`${ts}/tidal-drum-machines-alias.json`);
```
TodePond, line 43: what's the bet way to try out the repl? i only checked the web version works (with `pnpm dev`)

packages/superdough/superdough.mjs
```diff
@@ -17,11 +17,72 @@ import { loadBuffer } from './sampler.mjs';
...
+      if (typeof args[0] === 'string') {
+        return aliasBankPath(args[0]);
+      } else {
+        return aliasBankMap(args[0]);
+      }
+    case 2:
+      return aliasBankMap({ [args[0]]: args[1] });
+    default:
+      throw new Error('aliasMap expects 1 or 2 arguments, received ' + args.length);
+  }
```
TodePond, line 81: if theres a better place for this function to live lemme know and i'll move it

website/public/tidal-drum-machines-alias.json
```diff
@@ -0,0 +1,69 @@
...
+  "RolandS50": "S50",
+  "RolandSH09": "SH09",
+  "RolandSystem100": "System100",
+  "RolandTR505": "TR505",
+  "RolandTR606": "TR606",
+  "RolandTR626": "TR626",
+  "RolandTR707": "TR707",
+  "RolandTR727": "TR727",
+  "RolandTR808": "TR808",
+  "RolandTR909": "TR909",
```
TodePond, line 53: kinda wanna add "909" as another alias but its a slippery slope

website/public/tidal-drum-machines-alias.json
```diff
@@ -0,0 +1,69 @@
...
+  "SakataDPM48": "DPM48",
+  "SequentialCircuitsDrumtracks": "CircuitsDrumtracks",
+  "SequentialCircuitsTom": "CircuitsTom",
+  "SergeModular": "Serge",
```
TodePond, line 57: hmm maybe not
```suggestion
```

website/public/tidal-drum-machines-alias.json
```diff
@@ -0,0 +1,69 @@
...
+  "SimmonsSDS400": "SDS400",
+  "SimmonsSDS5": "SDS5",
+  "SoundmastersR88": "R88",
+  "UnivoxMicroRhythmer12": "MicroRhythmer12",
+  "ViscoSpaceDrum": "SpaceDrum",
```
TodePond, line 62: its real my neighbour told me

TodePond:
ready for review now!
Read the pull request description for the current changes.
left some comments too

packages/repl/prebake.mjs
felixroos: i test with this https://github.com/tidalcycles/strudel/blob/main/examples/buildless/web-component-no-iframe.html replacing the unpkg url with the built repl package

TodePond: can u give an example of what the contents of your `web-component-no-iframe.html` file are when you do this? could write it up in the meantime too

felixroos: i've added infos on testing to `packages/repl/README.md` . also tested it and the alias stuff works :)

packages/repl/prebake.mjs
felixroos: it very likely be available on https://strudel.cc/tidal-drum-machines-alias.json after merge/deploy

packages/superdough/superdough.mjs
felixroos: looks good

website/public/tidal-drum-machines-alias.json
felixroos: yeah i guess 2 more or less characters doesnt make a lot of difference..

packages/repl/prebake.mjs
felixroos: in theory, both github links wouldn't need to be fetched, instead the json could be used directly hmmm

felixroos:
> ready for review now! Read the pull request description for the current changes. left some comments too

looks all good to me! i think we could merge this then?

TodePond:
oh my god

PR #1246, opened 2025-01-18
added lowpass filter lfo

thckr:
added lowpass filter lfo

note("g1").s("sawtooth").lpf(400).lplfo("<1 2 3 4>)

### DIFF
--- a/packages/core/controls.mjs
+++ b/packages/core/controls.mjs
@@ -547,6 +547,19 @@ export const { cut } = registerControl('cut');
 export const { cutoff, ctf, lpf, lp } = registerControl(['cutoff', 'resonance', 'lpenv'], 'ctf', 'lpf', 'lp');
 
 /**
+ * Sets the lowpass filter lfo modulation depth.
+ * @name lplfo
+ * @param {number | Pattern} modulation depth of the lowpass filter lfo between 0 and _n_
+ * @example
+ * note("c2 e2 f2 g2")
+ * .sound('sawtooth')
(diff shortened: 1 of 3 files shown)


daslyfe:
We have a method for continuous modulation without using "real" lfo nodes. See the phaser for an example! The only thing holding back from implementing them everywhere is that we haven't discovered a great way yet to describe the modulation in strudel code to a general implementation

felixroos:
> We have a method for continuous modulation without using "real" lfo nodes. See the phaser for an example! The only thing holding back from implementing them everywhere is that we haven't discovered a great way yet to describe the modulation in strudel code to a general implementation

i think it would still be desirable to have a dedicated filter lfo control. we could refactor the definition later if we find a more general approach to lfos. to me it would be huge to be able to write .lplfo(4) and get that wobbly filter

daslyfe:
> > We have a method for continuous modulation without using "real" lfo nodes. See the phaser for an example! The only thing holding back from implementing them everywhere is that we haven't discovered a great way yet to describe the modulation in strudel code to a general implementation
> 
> i think it would still be desirable to have a dedicated filter lfo control. we could refactor the definition later if we find a more general approach to lfos. to me it would be huge to be able to write .lplfo(4) and get that wobbly filter

sounds good! we can use the lfo worklet for continuous modulation, and also use the same language we used for the phaser possibly? https://github.com/tidalcycles/strudel/blob/main/packages/superdough/superdough.mjs#L196C10-L196C19

Issue #595, opened 2023-06-11
"as" function


felixroos:
...which allows using colon notation for arbitrary combinations of controls:

```js
const as = register('as', (mapping, pat) => {
  return pat.fmap((v) => {
    // TODO: using aliases e.g. lpf does not work yet...
    v = Array.isArray(v) ? v : [v];
    v = Object.fromEntries(mapping.map((prop, i) => [prop, v[i]]));
    //console.log(v)
    return v;
  });
})

"[c2 a2 f2 e2]:{200 1000 800 2000}%3".as("note:<cutoff hcutoff>").s('sawtooth')
```

https://strudel.tidalcycles.org/?Z_qnuC-jOaL9

this would even allow to pattern the params itself:

https://strudel.tidalcycles.org/?5lh-8JXdPlKH

maybe this could also work for any pattern method and not just controls, similar to https://github.com/tidalcycles/strudel/issues/566 does it

felixroos:
also useful as tie replacement: https://strudel.cc/?svGlj81FALJ2 ref https://github.com/tidalcycles/strudel/issues/52

felixroos:
... or to set anchors per chord: https://strudel.cc/?l2-ExgNDGd8o

felixroos:
another variant by @daslyfe  https://strudel.cc/?cFh54k5oQlki

PR #1247, opened 2025-01-18
add "as" function + getControlName

felixroos:
adds as function:

```js
"c:.5 a:1 f:.25 e:.8".as("note:clip")
```

fixes https://github.com/tidalcycles/strudel/issues/595

- [x] add doc


yaxu:
Would it be possible to also support `"c:.5 a:1 f:.25 e:.8".as(note, clip)` ?

felixroos:
> Would it be possible to also support `"c:.5 a:1 f:.25 e:.8".as(note, clip)` ?

it might work if we say that multiarg is interpreted as a list of functions.. but not sure what a use case would be.

yaxu:
.. and a bit wackier, but following that thought, howabout `s("bd sd").apply("4 5".as("<fast hurry>"))` or `s("bd sd").as("<fast hurry>", "4 5")`

felixroos:
> .. and a bit wackier, but following that thought, howabout `s("bd sd").apply("4 5".as("<fast hurry>"))` or `s("bd sd").as("<fast hurry>", "4 5")`

hm yeah that wouldn't work, as it's only for controls, but i see the appeal..

yaxu:
> it might work if we say that multiarg is interpreted as a list of functions.. but not sure what a use case would be.

Yep I was thinking these would be called as methods, but true they need to call toplevel functions which is harder to implement.

yaxu:
> hm yeah that wouldn't work, as it's only for controls, but i see the appeal..

I guess it could be implemented as a string of method calls to give some experimental behaviour for 'free'..

felixroos:
any objections against merging this as is? making it work with more than controls surely is interesting but i guess that'd would require more brain power / effort / time to get it working. it would probably not produce a breaking change then

yaxu:
Yes agreed it can be built on later so lets merge!

Issue #1207, opened 2024-10-29
Cycles and pulses - maturing stepwise UI


yaxu:
The stepwise functions are useful but need some rationalisation. 

I propose:
* Rename all references of 'tactus' to 'pulse'. They're synonymous but the latter is less obscure.
  * `setTactus` could become `pulse` or maybe `pulseAt`.
  * `s_expand` could become `pulseBy`, as it does the same as `pulse` but with a factor change rather than absolute value
* Renaming `steps` to `cycle`. Then it should be slightly less ambiguous, as a function for changing hap density and therefore the number of pulses per cycle. Other potential options are `cycleAt`, or `recycle`.
  * The difference between above `pulse` and `cycle` is that `pulse` just sets the pulse, whereas `cycle` both sets the pulse and changes the pattern density (i.e. via `fast`) to match. 
* Rename `s_extend` to `cycleBy`, as it does the same as `steps` but with a factor change rather than absolute value for the tactus/pulse.

| old | new | hap density | pulse | absolute/factor | use case |
| -- | -- | -- | -- | -- | -- |
| fast | fast | x | | factor | Used when you want the pattern to be faster relative to the pulse. This is usually the case, otherwise in e.g. `sound(s_cat("bd sd".fast(2), "cp"))`, the `bd sd` would repeat twice but each hap would have the same duration as `cp` |
| steps | cycle | x | x | absolute | Used when you want to change the speed of a pattern to match a given 'pulse'. Useful e.g. where you have two patterns with different pulses and want to align them 'stepwise' so they have matching meters. |
| s_extend | cycleBy | x | x | factor | Used when you want the pattern and pulse to be increased along with the pattern speed. This is basically `!` in mininotation. |
| setTactus | pulse | | x | absolute | |
| s_expand | pulseBy | | x | factor | Increases (or decreases) the pulse density, thereby reducing (or increasing) the hap density per pulse. Equivalent to `@` in mininotation. |

yaxu:
Or given the use cases above..

| old | new | hap density | pulse | absolute/factor | mininotation equivalent |
| -- | -- | -- | -- | -- |
| fast | fast | x | | factor | `*` |
| steps | pulses | x | x | absolute | `{}%` |
| s_extend | repeat | x | x | factor | `!` |
| setTactus | | | x | absolute | |
| s_expand | expand | | x | factor | `@` |

yaxu:
[...]

PR #1262, opened 2025-01-24
Polish, rename, and document stepwise functions

yaxu:
WIP on #1207


yaxu:
Can't work out why the expand tests are failing..

yaxu:
I'll probably add backward compatibility on the names, even though they were 'experimental'..

yaxu:
Probably not the final word on this but I might merge it for now?

If/when https://github.com/tidalcycles/Tidal/pull/1111 is solved, I think things will become clearer..

yaxu:
Renaming `taper` to `shrink`, and adding `grow`, changing behaviour of negative values to match, so shrink is progressive drop, grow is progressive take. Combine amount and times into single list param, with times defaulting to the tactus.

Making `drop`  drop from the start for positive values and from the end for negative ones.

PR #1244, opened 2025-01-18
Add MIDI Program Change, SysEx, NRPN, PitchBend and AfterTouch Output 

nkymut:
This update adds support for sending MIDI Program Change and System Exclusive (SysEx) messages for @strudel/midi package. It builds upon the changes made in #1212; however, MIDI mapping and Hi-res CC have not been tested yet.

### Program Change:

```javascript
// [obsolete]note("c a f e").pc("<0 1 2>").midichan(1).midi()
$: note("c a f e").progNum("<0 1 2>").midichan(1).midi()
$: "10 20 30".progNum().midichan(1).midi()

// midi/gm.mjs provides mapping of gm intrrument names and pg numbers. 
// [obsolete] note("c a f e").pc(gm['marimba']).midichan(2).midi()
note("c a f e").progNum(gm['marimba']).midichan(2).midi()
//use midicmd
$: note("c a f e").midicmd("progNum:10").midichan(1).midi()
$: "progNum:20 progNum:30".midicmd().midichan(1).midi()
```

### SysEx Messages:

```javascript
// Send a simple SysEx message
let id = 0x43; //Yamaha ID for Gakken(Yamaha) NSX-39
let data = "0x79:0x09:0x11:0x0A:0x00:0x00"; // Set NSX-39 voice to say "Aa"

$: note("c d e f e d c").sysex([id,data]).midi();
$: note("c d e f e d c").sysexid(id).sysexdata(data).midi();

//ID can be an array of numbers
//let id = "0x00:0x20:0x32"; //Behringer 

```
### midibend && miditouch

`midibend` sets MIDI pitch bend (-1 - 1)
`miditouch` sets MIDI key after touch (0-1)

```javascript

$: note("c d e f e d c").midibend(sine.slow(4).range(-0.4,0.4)).midi();
$: note("c d e f e d c").miditouch(sine.slow(4).range(0,1)).midi();

```
### cc,progNum, sysex for `midicmd` Messages:
- `cc` - sends MIDI control change messages.
- `progNum` - sends MIDI program change messages.
- `sysex` - sends MIDI system exclusive messages.

```javascript
stack(
  // "cc:ccn:ccv"
  midicmd("cc:74:1").midi('IAC Driver'),
  // "progNum:progNum"
  midicmd("progNum:1").midi('IAC Driver'),
  // "sysex:[sysexid]:[sysexdata]"
[...]

PR #1274, opened 2025-02-02
midimaps

felixroos:
after discussion in https://github.com/tidalcycles/strudel/pull/1212 this PR adds support for defining midimaps:

```js
// defines a midimap for the given name
midmaps({
  miniv3: { 
    lpf: { ccn: 74, min:0, max:20000, exp:0.5 },
    gain: { ccn: 7, min:0, max:1 }
  }
})
// create helper function to use the midimap and the default midi device
register('miniv3', pat => pat.midimap('miniv3').midi())

// because it's midi, we can send stuff in parallel
$: note("[<g bb <d4!3 eb4>> -]*8").miniv3()
$: gain("<.25 .5 1>*8").miniv3()
$: lpf(sine.range(20,8000).slow(1)).segment(64).miniv3()
```

edit: i've also now added the function `defaultmidimap` which allows defining the default midimap:

```js
defaultmidimap({
  lpf: { ccn: 74, min:0, max:20000, exp:0.5 },
  gain: { ccn: 7, min:0, max:1 }
})
$: note("[<g bb <d4!3 eb4>> -]*8").midi() // no need to set midimap control
```

todo:

- [x] doc
- [x] load from url
- [x] test
- [x] gather feedback


Bubobubobubobubo:
It makes sense to also collect #1244 now, as it was attached to PR #1212. Other than that, very good design 😄.

PR #1284, opened 2025-02-09
Midimap+

nkymut:
This update enables midimaps to register control names that are not limited to Strudel's controlNames. 
This provides greater flexibility in customization while maintaining compatibility with existing mappings.
However, further consideration is needed regarding potential namespace pollution.

``` jsx
midimaps({ nsx: { gain: 7, 
                  pan:10, 
                  modulate:1, 
                  express:11, 
                  sustain:64, 
                  reverb:91,
                  chorus:93,
                  variation:94  } })
           
```

This PR includes updates from #1244


felixroos:
great, yes that was still missing from the midimap PR. so i guess we discuss / merge https://github.com/tidalcycles/strudel/pull/1244 first, before looking at this one?

nkymut:
> great, yes that was still missing from the midimap PR. so i guess we discuss / merge #1244 first, before looking at this one?

Yup, #1244 is good to go. And we can also look at if we map non-cc messages.

felixroos:
as #1244 is now (finally) merged, we could go with this next.. @nkymut could you maybe merge main into this one? it's a bit hard to see changes otherwise

nkymut:
Hi @felixroos thank you! I have merged the main.

packages/midi/midi.mjs
```diff
@@ -100,10 +100,23 @@ export const midicontrolMap = new Map();
...
+      // Convert number to object with ccn property
       if (typeof mapping === 'number') {
         mapping = { ccn: mapping };
       }
-      return [getControlName(key), mapping];
+
+      const controlName = getControlName(key);
+
+      // Register the control in the midicontrolMap if it doesn't exist in
+      if (!midicontrolMap.has(controlName)) {
```
felixroos, line 111: what if the control name exists already? then we shouldn't register it i think. when doing so, we might override it in a way that breaks it. for example, some controls are allowed to set multiple controls, like `cutoff:resonance:lpenv`, which would probably break when calling `registerControl('cutoff')`, as things like `note("g1").cutoff("200:8:4")` wouldn't set resonance and lpenv anymore

nkymut: Yes, that's what I meant by namespace pollution.
I think an easier solution would be to add a prefix like `mm_` to distinguish the names.
Alternatively, it could just throw an error if `controlAlias.has(controlName)` returns true, and prompt the user to choose a different control name.

felixroos: why not reuse existing names? one idea behind these midimaps was that you could reuse a non-mini pattern with your midimap without changing control names (e.g. lpf maps to your synths cutoff cc)

nkymut: I see, in that case how about registering a selective control function that checks whether `hap.value.midi` exists, and then decides whether to use the original lpf or the midi-mapped lpf? Or is there a better way to tell if the current trigger is to MIDI, SuperDough, or else?

nkymut:
Its been a while. 
As an interim fix, a warning is now shown when midimap overwrites a Strudel API method.

I've tried Throwing an error, but I find it too restrictive. For long-term flexibility—especially to allow smooth switching between MIDI and non-MIDI in the same code—it may be better to permit explicit destructive changes.

It would be more ideal if we could switch controlAlias based on output modes (midi, OSC, superDough) and active midimap, since the root cause is that all Pattern.prototypes are shared.

And currently It is hard to restore overwritten Strudel API methods to their default state beside refreshing the page.

PR #1287, opened 2025-02-18
Create Pattern Page Pagination 

daslyfe:
Also
- clip maxiumum title display size for a pattern in the patterns tab
- created input debounce utility
Screenshot 2025-02-18 at 10 42 45 AM

@matthewkaney it would be excellent if you could review the accessibility of the implementation here. I think I did it correctly but could use your experience


website/src/repl/components/panel/PatternsTab.jsx
```diff
@@ -84,107 +88,184 @@ function ActionButton({ children, onClick, label, labelIsHidden }) {
...
+        <PatternButtons
+          onClick={(id) => patternOnClick(id)}
+          started={context.started}
+          patterns={patterns}
+          activePattern={activePattern}
+        />
+      </div>
+      <div className="flex items-center gap-2 py-2">
+        <label htmlFor="pattern pagination">Page</label>
+        <Pagination id="pattern pagination" currPage={page} onPageChange={onPageChange} />
```
matthewkaney, line 193: This `id` isn't being used in the component currently. I think you should be passing it down and attaching it to the inner input tag that contains the page number, otherwise that field doesn't have an accessible label.

website/src/repl/components/incrementor/Incrementor.jsx
```diff
@@ -0,0 +1,57 @@
+import { cn } from 'tailwind_utils.mjs';
+import { Textbox } from '../textbox/Textbox';
+
+function IncButton({ children, label, className, ...buttonProps }) {
+  return (
+    <button
```
matthewkaney, line 6: Looking at [this example](https://www.w3.org/WAI/ARIA/apg/patterns/spinbutton/examples/datepicker-spinbuttons/), it looks like these buttons should be excluded from the Tab order with a `tabindex="-1"` attribute, since a keyboard user can already focus the main input component and edit the number with arrow keys.

They should still have labels though, for people who are using a screen reader with a touch screen. That said, I think it might be good to allow overriding the default increment/decrement labels with props, given that in this usage something like "next page" might be more descriptive than "increment."

website/src/repl/components/panel/Reference.jsx
```diff
@@ -28,12 +29,7 @@ export function Reference() {
     <div className="flex h-full w-full p-2 text-foreground overflow-hidden">
       <div className="h-full  flex flex-col gap-2 w-1/3 max-w-72 ">
         <div class="w-full flex">
-          <input
-            className="w-full p-1 bg-background rounded-md border-none"
-            placeholder="Search"
-            value={search}
-            onInput={(event) => setSearch(event.target.value)}
-          />
+          <Textbox placeholder="Search" value={search} onChange={setSearch} />
```
matthewkaney, line 32: It looks like this is messing with the styling of the reference search field, since it was previously relying on the `w-full` class in order to not overlap with the other column of the reference page.

matthewkaney (changes requested):
Sorry it took me a couple of days on this! Overall, it looks good—I flagged the couple of things that I noticed.

website/src/repl/components/panel/Reference.jsx
daslyfe: updated! ty

website/src/repl/components/panel/PatternsTab.jsx
daslyfe: Good callout, fixed!

website/src/repl/components/incrementor/Incrementor.jsx
daslyfe: incorperated both suggestions, thanks!

daslyfe:
Updated and incorporated all suggestions, thanks for reviewing @matthewkaney !

PR #1294, opened 2025-02-23
Initial random value at 0 for `rand` and other dependencies of `timeToRand`

yaxu:
This will change some tunes..


felixroos:
makes sense to me!

daslyfe:
I have been teaching people locally about using constrained randomness for melodies and it has become a common shared technique, ex: `note(irand(24).ribbon(0,1))’ so I am concerned that this would change a lot of people’s songs in an undesired way.

PR #1304, opened 2025-03-04
feat: Create Pulse Oscillator  with variable PWM

daslyfe:
TODO: add pulse width modulation lfo


felixroos:
just noticed this isnt documented yet.. would be nice for the release notes

PR #1305, opened 2025-03-06
Add ? and _ operators to mini notation page

makmanalp:
Filling in missing operators that Strudel already has.

Context: https://discord.com/channels/779427371270275082/1191011773311107153/1346812252648378469

### DIFF
--- a/website/src/pages/learn/mini-notation.mdx
+++ b/website/src/pages/learn/mini-notation.mdx
@@ -156,7 +156,11 @@ But to play multiple chords in a sequence, we have to wrap them in brackets:
 
 ## Elongation
 
-With the "@" symbol, we can specify temporal "weight" of a sequence child:
+With the "_" symbol, we can manually extend the duration of an event:
+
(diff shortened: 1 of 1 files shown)


yaxu:
I think at this point `_` and `@` are aliases, the difference being that `@` isn't valid in identifiers so `bd_3` doesn't work, but `b _3` does (being the same as `b@3`)

felixroos:
i wonder if we need the alias ? maybe we can not tell people about "_" to potentially reuse it later for something else

yaxu:
quietly deprecating sounds good..

website/src/pages/learn/mini-notation.mdx
```diff
@@ -156,7 +156,11 @@ But to play multiple chords in a sequence, we have to wrap them in brackets:
 
 ## Elongation
 
-With the "@" symbol, we can specify temporal "weight" of a sequence child:
+With the "\_" symbol, we can manually extend the duration of an event:
```
felixroos, line 159: maybe leave that out, as we try to quietly deprecate "_", in favor of "@" ?

felixroos:
@makmanalp would you want to remove the "_" part? the rest looks good to me

Issue #1228, opened 2024-12-23
autoclip


felixroos:
<img width="594" alt="image" src="https://github.com/user-attachments/assets/d11e2feb-4beb-4c71-a4ed-6a637f3f9eb9" />

this would prevent a good chunk of crashes / glitches

yaxu:
Do the piano samples need to be that long? It might just be enough to trim all the 'default' samples..

felixroos:
> Do the piano samples need to be that long? It might just be enough to trim all the 'default' samples..

they could be trimmed i guess.. generally, most sample banks have very long tails, so this will be a problem for other samples too.
not clipping long samples makes it also very easy to stack many instances by accident, e.g. s("sax") . when the sample is very long, it can take a lot of time until the "stack of doom" has finished playing. imo having clip in by default for non percussive samples (>1s) would be a more sensible default.

felixroos:
example usage:

```js
autoclip(0) // automatically clips all samples to hap length
autoclip(1) // automatically clips all samples >=1 s to hap length
autoclip(Infinity) // samples are never clipped automatically

$: s("sax") 
```

yaxu:
Looking at the piano samples some are over 24 seconds. That seems really long to me but I have not spent much time with pro sample packs..
I think it could be confusing to have samples clipped automatically, e.g. when working with drum loops or stems. Could the default clip info go into the sample pack metadata?

felixroos:
what i mean with "sonic use", i mean the difference between:

```js
note("c a f e").s("piano")
// and
note("c a f e").s("piano").clip(1).rel(4)
```

is marginal.. in the first case, you'd get 24s samples, in the second you'd get 5s samples. the idea is that if you want to have cpu costly behavior, you have to opt in instead of opt out, because people will either don't know about it or forget it

felixroos:
one drawback is that things will sound a bit worse with the default settings, as you won't get the floatyness of layered samples by default. but that is already the case with synths anyway..

PR #1313, opened 2025-03-22
feat: clip samples to duration when the buffer is longer than 1.5 seconds

daslyfe:
This is one of the biggest problems I see people running into when using long samples without clip() causes performance drops and crackles. Now percussion samples (usually shorter than 1.5 seconds) can keep their tales and longer sounds will be the appropriate duration.

### DIFF
--- a/packages/superdough/sampler.mjs
+++ b/packages/superdough/sampler.mjs
@@ -303,10 +303,9 @@ export async function onTriggerSample(t, value, onended, bank, resolveUrl) {
   const sbs = await getSampleBufferSource(value, bank, resolveUrl);
  const { bufferSource, sliceDuration, offset,bufferDuration } = sbs
 
-  // asny stuff above took too long?
+  // any stuff above took too long?
   if (ac.currentTime > t) {
     logger(`[sampler] still loading sound "${s}:${n}"`, 'highlight');
-    // console.warn('sample still loading:', s, n);
     return;
   }
   if (!bufferSource) {
--- a/packages/superdough/sampler.mjs
+++ b/packages/superdough/sampler.mjs
@@ -301,7 +301,7 @@ export async function onTriggerSample(t, value, onended, bank, resolveUrl) {
   let [attack, decay, sustain, release] = getADSRValues([value.attack, value.decay, value.sustain, value.release]);
 
   const sbs = await getSampleBufferSource(value, bank, resolveUrl);
- const { bufferSource, sliceDuration, offset,bufferDuration } = sbs
+  const { bufferSource, sliceDuration, offset, bufferDuration } = sbs;
 
   // any stuff above took too long?
   if (ac.currentTime > t) {
@@ -323,7 +323,7 @@ export async function onTriggerSample(t, value, onended, bank, resolveUrl) {
   const envGain = ac.createGain();
   const node = bufferSource.connect(envGain);
 
-  const clipDeltaSeconds = 1.5
+  const clipDeltaSeconds = 1.5;
 
   // if none of these controls is set, the duration of the sound will be set to the duration of the sample slice
   if (clip == null && loop == null && value.release == null && bufferDuration < clipDeltaSeconds) {
--- a/packages/superdough/sampler.mjs
+++ b/packages/superdough/sampler.mjs
@@ -300,7 +300,8 @@ export async function onTriggerSample(t, value, onended, bank, resolveUrl) {
   // destructure adsr here, because the default should be different for synths and samples
   let [attack, decay, sustain, release] = getADSRValues([value.attack, value.decay, value.sustain, value.release]);
 
-  const { bufferSource, sliceDuration, offset } = await getSampleBufferSource(value, bank, resolveUrl);
+  const sbs = await getSampleBufferSource(value, bank, resolveUrl);
+ const { bufferSource, sliceDuration, offset,bufferDuration } = sbs
 
   // asny stuff above took too long?
   if (ac.currentTime > t) {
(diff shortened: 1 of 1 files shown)


felixroos:
nice! i think it would be good to have a function to define this limit, like discussed here: https://github.com/tidalcycles/strudel/issues/1228 but 1.5s seems to be a good default

daslyfe:
Sounds like a good idea to me. Should the function live in the settings.mjs file? How would I make it callable from the repl window?

daslyfe:
> I'm a bit unsure about this, I think it will surprise people in some cases if it's default on.
> 
> The problem could also be solved by one or both of
> 
> * trimming the default samples to be more sensible
> * a max polyphony option that stops long running notes (or some other scheme)
> 
> Saying that I do sometimes wonder what would happen if samples were all clipped to hap duration, like synths. Probably mininotation would have evolved differently to cope/take advantage of that, to make it easier to create overlapping haps.

I would strongly advocate for this solution or "all clipped to hap duration". I did a guest lecture for the berklee experimental music program recently where the students had been learning generative music concepts with strudel.  Nearly all of them had issues with their projects crashing due to too many buffers going on. I also saw the same thing happen to a girl who was performing this weekend.

yaxu:
> I would strongly advocate for this solution or "all clipped to hap duration" + using release time to get longer sample lengths.

When you say 'this solution' do you mean this PR or something I mentioned?

daslyfe:
> Yep agreed this is a super important issue to solve.
> 
> > I would strongly advocate for this solution or "all clipped to hap duration" + using release time to get longer sample lengths.
> 
> When you say 'this solution' do you mean this PR or something I mentioned?
> 
> I've done quite a lot of workshops and haven't really hit a performance issue, including at recent workshops with school kids who were pushing strudel _hard_ on chromebooks. I am used to doing workshops with dirt samples rather than using the longer piano sounds etc though.
> 
> Clipping sounds at a set limit does feel drastic to me, and clipping them at hap duration would change the quality of music a lot.. So I am still in favour of solving this by stopping sounds when a configurable max polyphony is reached, but don't know how tricky that would be to implement in superdough. The classic dirt just stops the longest running sound when it hits the max, as far as I remember.

I was referring to this PR. Unfortunately I have seen this issue happen to multiple performers. I also saw a performer's set glitch out and freeze SXSW this year because of it as well, so I think it would be good if we can solve this sooner than later imo. 


Another idea I had, which might be a more ideal solution, is to add a property for clip value for a sound in the strudel JSON. This way you can specify that your drum samples have undefined clip behavior, and everything else has a clip of 1 by default so the clip behavior is consistent with the synth sounds.

yaxu:
I'm not totally against the idea of having clip default to 1 to make sounds consistent with synths. Maybe `.clip(-1)` or `.noclip()` could turn it off. It seems a big change tho! 

I guess unbound polyphony will always cause freezes - even if sounds are clipped, too many could still be triggered by mistake. I've not looked into superdough code yet..

felixroos:
i think max voice count is the way to go to fix glitches related to too much layering. (like cut, but polyphonic, fifo). not sure if autoclipping by default is best, but surely it would be good to have as an option

yaxu:
I remember a synth where you could set how notes were allocated. It took me a while to find the search term to use - 'note priority' or 'stealing priority', .[e.g](https://www.plogue.com/bidule/help/ch08s03.html).:

* Arbitrary: Don't care, the adapter will randomly steal one
* Quietest: The quietest note will go first (this is usually less noticeable when many notes are playing)
* Closest to New: This will steal the note thats the closest to the new note you want to play
* Oldest: Will steal the note that's been playing for the longest period of time
* Newest: Will steal the note that's been playing for the shortest period of time

Monosynths seem to generally use lowest priority.

I think newest priority is probably the best we can do when we don't really know anything about the sounds that are being triggered though.

felixroos:
> I remember a synth where you could set how notes were allocated. It took me a while to find the search term to use - 'note priority' or 'stealing priority', .[e.g](https://www.plogue.com/bidule/help/ch08s03.html).:
> 
>     * Arbitrary: Don't care, the adapter will randomly steal one
> 
>     * Quietest: The quietest note will go first (this is usually less noticeable when many notes are playing)
> 
>     * Closest to New: This will steal the note thats the closest to the new note you want to play
> 
>     * Oldest: Will steal the note that's been playing for the longest period of time
> 
>     * Newest: Will steal the note that's been playing for the shortest period of time
> 
> 
> Monosynths seem to generally use lowest priority.
> 
> I think newest priority is probably the best we can do when we don't really know anything about the sounds that are being triggered though.
> 
> Maybe it could be nice to have a priority system on top so `.priority(-100)` would be stopped before any `.priority(100)` haps.. Comparing priority first then age

thats a good list! id expect oldest to be the default

felixroos:
we now have https://github.com/tidalcycles/strudel/pull/1317 , so i wonder how this changes the clip discussion. it might still be desirable to not have samples that are very long build up a ton of overlaps. max polyphony is 128, so 128 overlaps can still happen

yaxu:
I'm still in the "don't do that then" camp, trimming silence off the default samples, rather than trimming them on playback. This would avoid unnecessary bandwidth use and download time of all that unwanted silence/near silence. It also means if people really want super long tails on their samples they can do that without having to remember to turn of automatic clipping.

yaxu:
> thats a good list! id expect oldest to be the default

Agreed, confusingly 'oldest' in that list is 'newest priority'

daslyfe:
Adding a release time would also make the samples longer, you don’t have to mess with the clip value for that. Currently the voice stealing happens so the longest playing voice gets stolen

PR #1314, opened 2025-03-24
small feat: Add alias for segment and ribbon

daslyfe:
Also: updated related documentation and added more examples


yaxu:
seg / rib fine by me!

PR #1317, opened 2025-03-30
feat: add max polyphony feature for superdough

daslyfe:
Setting panel option added

### DIFF
--- a/packages/superdough/superdough.mjs
+++ b/packages/superdough/superdough.mjs
@@ -374,6 +374,9 @@ export function resetGlobalEffects() {
   analysersData = {};
 }
 
+let allAudioNodeChains = [];
+let maxPolyphony = 12;
+
 export const superdough = async (value, t, hapDuration) => {
   const ac = getAudioContext();
   t = typeof t === 'string' && t.startsWith('=') ? Number(t.slice(1)) : ac.currentTime + t;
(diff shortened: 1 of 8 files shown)


packages/superdough/superdough.mjs
```diff
@@ -473,13 +480,29 @@ export const superdough = async (value, t, hapDuration) => {
 
   gain = nanFallback(gain, 1);
 
+  const chainID = Math.round(Math.random() * 10000);
```
felixroos, line 483: maybe id can be an ever increasing integer modulo maxPolyphony? 10000 seems like occasional collisions might happen (might be very unlikely)

packages/superdough/superdough.mjs
```diff
@@ -473,13 +480,29 @@ export const superdough = async (value, t, hapDuration) => {
...
+
+  // oldest audio nodes will be removed if maximum polyphony is exceeded
+  for (let i = 0; i <= allAudioNodeChains.size - maxPolyphony; i++) {
+    const ch = allAudioNodeChains.entries().next();
+    const key = ch.value[0];
+    const chain = ch.value[1];
+    if (key == null) {
+      continue;
+    }
+    chain?.forEach((node) => node?.disconnect());
```
felixroos, line 493: might need to fade out to avoid clicks when the sound is still audible?

felixroos:
great stuff! i wonder if 128 is large enough but maybe it is

packages/superdough/superdough.mjs
daslyfe: Good idea, I reworked some things to make this possible

packages/superdough/superdough.mjs
daslyfe: Im concerned that async behavior could cause clashes with modulo id, Increased the multiplier to make it very very unlikely that a clash would happen.

daslyfe:
> great stuff! i wonder if 128 is large enough but maybe it is

It seems to be a good default for performance purposes imo

yaxu:
What's a good way to push the polyphony? I'm struggling to hear different behaviour between main and this PR on firefox/linux on my thinkpad x1. I'm currently testing it with a breakbeat * 1000 and it's kind of OK on both...

daslyfe:
It’s hard to hear the difference because it fades out the cut voices when
there are so many, if you lower the polyphony to like 12 and use the piano
sample you can start to hear it more

daslyfe:
If you are talking about pushing the cpu though, using the piano sample, or
any of the audioworklet effects or sources like “supersaw” with long
release times  you can watch your cpu go up with high polyphony

yaxu:
Gotcha thanks. Yep this is working well for me - piano destroys strudel main but voicemax recovers straight away. 

I haven't done detailed tests but couldn't overload strudel in the same way when I tried with a breakbeat sample. Those piano sounds must be huge..

PR #1319, opened 2025-04-03
fix: replace empty spaces in registered sound keys

daslyfe:
`my cool   sound.wav` becomes `my_cool_sound` when registered in the sound panel, allowing you to pattern it properly


felixroos:
good idea

PR #1329, opened 2025-04-24
Add new "import-sounds" tab with explanation on folder import

hpunq:
Replaced the "import sounds" button with a tab that offers the same functionality + an explanation. This has been done to remove the ambiguity related to adding personal samples to strudel for beginners.

A little unsure of the implementation personally; it might make sense to refactor the tab text and ui into a new component, rather than keeping it within SoundsTab.

Either way, hopefully this is an improvement over the previous implementation and makes getting started a bit easier.


daslyfe:
Looks great! thank you. It would be great if you could update the DOCS as well in a seperate PR to accoutn for the new workflow

PR #1330, opened 2025-04-26
Prepare

### LINKED ISSUE #187: sample preloading query loop
Currently, samples are lazily loaded when they are first needed. for most samples, this leads to unwanted silence, because the loading takes longer than the 100ms latency.
It would be pretty useful if there was a separate query loop that preloads the samples with a higher latency, say x seconds.
To ensure there is always at least a buffer of x seconds between preloading and playback, the initial play action would then need to query the first x*2 seconds. Then, after x seconds, the next x seconds are loaded and so on.

prezmop:
resolves #187

This was annoying me so much that i took it upon myself to write it.

### DIFF
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -96,7 +96,7 @@ export class Cyclist {
           (phase, duration, _, t) => {
             try {
               const start = Math.max(t, this.preparedUntil);
-              end = t + this.prepareTime;
+              const end = t + this.prepareTime;
               this.preparedUntil = end;
 
               const haps = this.pattern.queryArc(start, end, { _cps: 1 });
--- a/packages/core/cyclist.mjs
+++ b/packages/core/cyclist.mjs
@@ -90,30 +90,32 @@ export class Cyclist {
       clearInterval,
     );
 
-    onPrepare? this.prepClock = createClock(
-      getTime,
-      (phase, duration, _, t) => {
-        try {
-          const start = Math.max(t, this.preparedUntil);
-          end = t + this.prepareTime;
-          this.preparedUntil = end;
+    onPrepare
+      ? (this.prepClock = createClock(
+          getTime,
+          (phase, duration, _, t) => {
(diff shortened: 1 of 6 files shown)


prezmop:
Currently I have not written any docs for the new functionality.
I have only written bare minimum code to implement the required functionality into `superdough` and added some code to call it in `Cyclist` (`NeoCyclist` remains untouched).

yaxu:
Hey thanks for working on this, could you give an example for how `prepare()` is used?

prezmop:
Hey,
like `superdough()`, `prepare()` is called with a value containing all the sound properties. However, it does not need a deadline or duration, since it doesn't actually play the sound.
It can optionally be called before the sound is actually needed, so that it can be downloaded.

The following examples both assume that the `"sd"` sound has not been loaded yet.

 without prepare (this still works):
```js
superdough({s:"sd"},0)
// sound doesn't play
```
with prepare:
```js
prepare({s:"sd"})

// some time later

superdough({s:"sd"},0)
//sound plays
```

prezmop:
Note that it is up to whatever/whoever is using superdough to inform what sounds will be needed

felixroos:
i like this idea! i cant review the code until in 2 weeks tho.. i think this would already be a good addition even without automatically calling it from a scheduler. is this a promise, so can we await it before playback?

prezmop:
for a strudel example:
```js
$: n("<0 1 2 3 4 5 6 7>").s("kawai").scale("c:major")
```

Though this actually seems to work worse at the moment.
Looking at the network requests for the sounds they seem to take really long time.
Maybe the amount of parallel fetches needs to be limited.

prezmop:
Admittedly, i don't know much about js

prezmop:
> i like this idea! i cant review the code until in 2 weeks tho.. i think this would already be a good addition even without automatically calling it from a scheduler. is this a promise, so can we await it before playback?

Yes, it can be awaited

PR #1332, opened 2025-04-27
update docs to reflect import sounds tab change

hpunq:
Made a small few changes in the docs to reflect the new import sounds folder UI.


daslyfe:
Thanks!

PR #1334, opened 2025-04-27
FIX: Set default gain to 1

daslyfe:
A default gain of 0.8 is not intuitive, because it is an invisible value. This change allows you to think of 1 as full volume, 0.8 as 80%, 0.5 as 50% etc.


yaxu:
Agreed, `1` is default in tidal and until now I assumed that was the case in strudel too

felixroos:
yep also agree.. i'd prefer to do this breaking change in the next version, so we can finally release the current things without breaking too much

PR #1341, opened 2025-05-03
feat: delaytime in cycles

daslyfe:
allows relative delay to stay the same during tempo changes
defaults to dotted eight notes
easily get more rhythmic results with the delay


yaxu:
In superdirt default is seconds, but if you add the parameter `# lock 1`, it's calculated as cycles. Maybe it's nice to copy that behaviour?

daslyfe:
What if we had delaytime and delaysync(3/16) might be more discoverable

PR #1343, opened 2025-05-06
feat: Create pulsewidth (pw) and pulsewidth lfo parameters

daslyfe:
The LFO is a simple triangle oscillator with `pwsweep` and `pwrate` controls that uses the psuedo-continuous modulation approach I implemented for the phaser


felixroos:
nice! does `pw` replace `n` as a control to set the pulse width?

daslyfe:
> nice! does `pw` replace `n` as a control to set the pulse width?

currently the priority is pw ?? n to control the pulsewidth.

Issue #823, opened 2023-11-25
find a good multichannel / orbit / pattern abstraction


felixroos:
find a abstraction that allows using a second stereo output as a cue output

here's 2 graphics made bx @daslyfe 

<img width="891" alt="Screenshot_2023-11-21_at_4 18 21_PM" src="https://github.com/tidalcycles/strudel/assets/12023032/675153d1-7a57-4342-9639-6b2b5389a6ca"/>

<img width="686" alt="Screenshot_2023-11-21_at_4 29 08_PM" src="https://github.com/tidalcycles/strudel/assets/12023032/18159e18-2554-4dee-9444-7d71d92057cb"/>

channels are implemented in https://github.com/tidalcycles/strudel/pull/820

there were some more ideas in chat that I want to drop here:

- maybe at the top level you can set the q output? something like `setQ([2,3])`
- maybe use 1,2 as master out by default and 3,4 as cue by default if it exists
- it could default to the second stereo pair on the interface
- still requires some more thought about the relation between pattern names (like x in p(x) or 1...9 in p1..p9) and channels and orbits. also how this would work without using named patterns. I kind of like writing a single stack of patterns as opposed to individual blocks with .pX because it's easier to add transformations to multiple patterns at once / groups of patterns. the all function only partly solves this problem as you still cannot transform only a subset of patterns 
i also wonder if there would be a use case for being able to use the same orbit for different channels. or the same channel for different orbits
- I was thinking they would act more as traditional fx sends in a DAW, maybe you can configure/ set what channles they output to like `orbit(1).setChannels([3,4])`
kinda like `setCps`
- it seems like the channel should receive the signal without global effects then (aka pre-send / pre-orbit). post-send/orbit can potentially contain effects of other channels then
- the question is how would you route the effects into a channel..
the alternative would be to connect the orbit to the channel (post-send/orbit) . then either let the user set the channels and orbits manually or automatically set the orbit to the channel number
or do both with something like `.preCh` and `.postCh`
- here is how I was thinking it could work: `patternA = sound("bd").channels([2,3]).room(1).orbit(1)`
- `orbit(1).setChannels([5,6])`
- channels 5/6 would receive a completely wet output of the reverb signal, and channels 2/3 would receive a dry signal
[...]

PR #1344, opened 2025-05-07
feat: Multi Channel Orbits

daslyfe:
Multi Channel orbits can be enabled in the settings page. Ex: `.orbit(2) ` will route the orbits to channels 3 and 4 on the audio device if those channels are available.

Also: 
-Channels will now wrap around if a higher channel count than is available is selected, instead of clamping to the highest channel increment available.

### DIFF
--- a/packages/superdough/superdough.mjs
+++ b/packages/superdough/superdough.mjs
@@ -441,6 +441,11 @@ export function resetGlobalEffects() {
 }
 
 let activeSoundSources = new Map();
+//music programs/audio gear usually increments inputs/outputs from 1, we need to subtract 1 from the input because the webaudio API channels start at 0
+
+function mapChannelNumbers(channels) {
+  return (Array.isArray(channels) ? channels : [value.channels]).map((ch) => ch - 1);
+}
 
 export const superdough = async (value, t, hapDuration) => {
   const ac = getAudioContext();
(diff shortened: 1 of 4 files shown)


felixroos:
i haven't read all the changes, but i wonder what's the difference between this and the (undocumented) `channels` param? iirc you can do `s("bd").channels("3:4")` already. there's also this issue which seems related: https://github.com/tidalcycles/strudel/issues/823

daslyfe:
> i haven't read all the changes, but i wonder what's the difference between this and the (undocumented) `channels` param? iirc you can do `s("bd").channels("3:4")` already. there's also this issue which seems related: #823

The main functional difference is that the fx in the orbits themselves get assigned to the corresponding channels. Ex: `orbit(3)` will route the sound AND the fx in orbit 3 to channels 5 and 6 with multichannel orbits enabled. This is also the way people do multichannel audio in Tidal, and people who I have talked to expect the same behavior when coming over to Strudel :) Both ways of assigning channels are useful imo for different reasons

daslyfe:
you can also route the fx sends to different tracks by doing `s("sawtooth").orbit(2).ch("5:6")`  the fx will be sent to the orbits on channels "3+4" and the synth sound will go to channels "5+6"

daslyfe:
> #823

I think this would probably close out the scope of this issue

Issue #1349, opened 2025-05-16
autocompletion shows irrelevant comments.


nachitodev:
during my code writing I could notice that having the "**_auto-completion_**" option activated, some sections and comments are mixed, badly formatted or erroneously included by js-doc.
I attach some examples of what I described.

![Image](https://github.com/user-attachments/assets/e1379f33-b66f-4efc-859e-45c363102d55) #

PR #1350, opened 2025-05-16
fix: typo on docs causing problems with autocompletion.

nachitodev:
fixing some typos comments and things from #1349.


yaxu:
Thanks!

PR #1354, opened 2025-05-20
feat: Byte Beats!

daslyfe:
@emptyflash and I implemented his bytebeat algorithm at the Livecode London meetup today!  `s("bytebeat").bbexpr('t*(t>>15^t>>66)')`


felixroos:
seeing this, i think it would be a really low hanging fruit to *just* try and run kabelsalat in a short lived worklet

daslyfe:
That would be awesome! You should try it!

PR #1362, opened 2025-06-01
supradough poc

felixroos:
this is a poc for implementing a small part of superdough without ties to webaudio. it can be used like this:

```js
$: note("c eb g bb").add(note(12))
  .lpf(sine.range(.6,.7).slow(2)).lpq(0.2)
  .s("sawtooth")
  .dec(.2).att(.02)
  .delay(.5)
  .supradough()
```

the `supradough` method completely bypasses superdough and spawns an instance of the `dough-processor` worklet for each event. for each worklet, an instance of the `Dough` class is created, given the params for the event. 
the `Dough` class is self contained, so it could be used to render a strudel pattern in any environment that supports js (e.g. node.js) + it would be easy to prerender a strudel tune from a pattern, without the use of the web audio api. the `dough.mjs` contains ugens from kabelsalat. we don't really need a graph, as superdough is static.

### DIFF
--- a/packages/superdough/dough.mjs
+++ b/packages/superdough/dough.mjs
@@ -1,0 +1,445 @@
+// this is dough, the superdough without dependencies
+
+const ISR = 1 / sampleRate;
+// https://garten.salat.dev/audio-DSP/oscillators.html
+export class SineOsc {
+  phase = 0;
+  update(freq) {
+    const value = Math.sin(this.phase * 2 * Math.PI);
+    this.phase = (this.phase + freq / sampleRate) % 1;
+    return value;
+  }
+}
+
+export class ZawOsc {
+  phase = 0;
+  update(freq) {
+    this.phase += ISR * freq;
+    return (this.phase % 1) * 2 - 1;
+  }
+}
+
+function polyBlep(t, dt) {
+  // 0 <= t < 1
+  if (t < dt) {
+    t /= dt;
+    // 2 * (t - t^2/2 - 0.5)
+    return t + t - t * t - 1;
+  }
+  // -1 < t < 0
+  if (t > 1 - dt) {
+    t = (t - 1) / dt;
+    // 2 * (t^2/2 + t + 0.5)
+    return t * t + t + t + 1;
+  }
+  // 0 otherwise
+  return 0;
+}
+export class SawOsc {
+  //phase = Math.random();
+  phase = 0;
+  update(freq) {
+    const dt = freq / sampleRate;
+    let p = polyBlep(this.phase, dt);
+    let s = 2 * this.phase - 1 - p;
+    this.phase += dt;
+    if (this.phase > 1) {
+      this.phase -= 1;
+    }
+    return s;
+  }
+}
+
+export class TriOsc {
+  phase = 0;
+  update(freq) {
+    this.phase += ISR * freq;
+    let phase = this.phase % 1;
+    let value = phase < 0.5 ? 2 * phase : 1 - 2 * (phase - 0.5);
+    return value * 2 - 1;
+  }
+}
+
+export class Lpf {
+  s0 = 0;
+  s1 = 0;
+  update(s, cutoff, resonance = 0) {
+    // Out of bound values can produce NaNs
+    cutoff = Math.min(cutoff, 1);
+    resonance = Math.max(resonance, 0);
+    var c = Math.pow(0.5, (1 - cutoff) / 0.125);
+    var r = Math.pow(0.5, (resonance + 0.125) / 0.125);
+    var mrc = 1 - r * c;
+    var v0 = this.s0;
+    var v1 = this.s1;
+    // Apply the filter to the sample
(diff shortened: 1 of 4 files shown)


daslyfe:
this looks excellent, I will try it out tomorrow morning

daslyfe:
Works pretty well! I noticed that with long release times it ends in a popping sound, `$: note("c eb g bb").add(note(12))
  .lpf(sine.range(600,7000).slow(2)).lpq(0.2)
  .s("sawtooth")
  .att(.3).release(1)
  // .delay(.5)
  .supradough()` volume could also be lowered to match superdough volume. The filter sounds really tasty

daslyfe:
fixed the release envelope, volume, and clicking issue :)

felixroos:
> fixed the release envelope, volume, and clicking issue :)

nice! i failed to find the bug with my sleep deprived brain on the way back :D 

i still have question marks on how to use this tactic to implement the rest of superdough, mainly:

1. global fx
2. samples
3. ??

i see 2 potential ways:

1. spawn only a single worklet and send haps as postMessages + do voice allocation in dough
2. keep spawning one worklet per hap, implement fx in separate long living worklet + wire them together

imo 1 would be the "purest" approach, making it maximally portable. not sure what's more performant, but i guess we have to try

felixroos:
image

web audio glue is now super minimal :)

felixroos:
i've added another poc (dough-export.mjs) to show how to use Dough in a node.js script to render the audio for a pattern

packages/superdough/superdough.mjs
```diff
@@ -190,11 +190,18 @@ export function getAudioContextCurrentTime() {
   return getAudioContext().currentTime;
 }
 
+let externalWorklets = [];
+export function registerWorklet(url) {
```
felixroos, line 194: i've added this to be able to register worklets from the outside without depending on them from superdough

felixroos:
rudimentary feedback delay is now in! new demo:

```js
$: note("g1,bb1,d2,f2").struct("x(3,8)")
  .add(note("<12 5>/2")).add(note(12))
  .lpf(sine.range(80,200).slow(7))
  .lpenv(4).lpd(.2).lpq(0.2).dec(.2)
  .s("sawtooth")
  .jux(rev)
  .sometimes(add(note(12)))
  .dist(.1)
  .delay(.5)
  .postgain(.5)
  .supradough()
```

it's also interesting to compare with superdough, by commenting out the last line.
i think the filter sounds much better

daslyfe:
Can we make the filter and volume envelopes be exponential?

felixroos:
still a bit of refactoring to do, to make sure stereo samples (and eventually stereo waveforms) are processed accordingly.. all the local effects ugens need to exist once for every channel i guess. envelopes are only needed once. also not sure how to best integrate the existing sample map loading logic

packages/supradough/dough.mjs
```diff
@@ -234,11 +234,11 @@ export class ADSR {
           this.startTime = curTime;
           return 1;
         }
-        return lerp(time / attack, this.startVal, 1, this.curve);
+        return lerp(time / attack, this.startVal, 1, 1);
       }
       case 'decay': {
         let time = curTime - this.startTime;
-        let curVal = lerp(time / decay, 1, susVal, -this.curve);
+        let curVal = lerp(time / decay, 1, susVal, -this.decayCurve);
```
felixroos, line 241: curious: why only decay and not attack / release ?

daslyfe: It should be on the release and decay stage, the reason is that seems to be standard on most synthesizers so I played around with different curves and preferred the standard linear rate on attack, it was not scientific

felixroos:
> still a bit of refactoring to do, to make sure stereo samples (and eventually stereo waveforms) are processed accordingly.. all the local effects ugens need to exist once for every channel i guess. envelopes are only needed once.

fx now work channelwise! so stereo samples should now work

daslyfe:
Loading all samples into memory is one of the problem areas in superdirt
that causes initialization to be slow, if we can stream from disk somehow
or only load samples being used that would probs be better

packages/supradough/dough.mjs
```diff
@@ -59,6 +60,59 @@ export class SawOsc {
...
+
+      this.phase[n] += dt;
+      if (this.phase[n] > 1) {
+        this.phase[n] -= 1;
+      }
+    }
+
+    return sl + sr
+    //TODO: make stereo
+    // return [sl, sr];
```
felixroos, line 112: i'd go for a property no the class, rather than creating an array for each sample (avoids memory allocation)

daslyfe:
Added delayspeed to create pitched delay effects, implementation inspired by [this article  ](https://www.eventideaudio.com/50th-flashback-4-2-h910-harmonizer-the-product/)

`note("d d a# a".fast(2)).s("sawtooth").delay(.8).delaytime(1/2).delayspeed("<2 .5 -1 -2>")
.supradough()`

PR #1365, opened 2025-06-03
feat(sounds panel UI): group sounds by folder and allow starring favorites (interface)

laema9:
Hey! Here are some ideas for improvements to the "Sounds" panel interface:

I found the current sounds panel a bit hard to navigate, especially when dealing with lots of custom sounds — everything was in one long list, which made it tricky to find or organize things. I tried to improve that by grouping sounds into folders and adding the ability to star favorites ⭐ for quicker access. Next, I’d love to take it a step further by showing and previewing each individual sound inside the folders, so instead of just seeing a total count, you can easily listen to each one.

Before: 

strudel_sounds_panel_after

After: 

strudel_sounds_panel_after

I think this section can still be improved a lot! What do you think?


laema9:
I'm working on showing each individual sample inside every folder, so it's much easier to preview them and copy/paste directly into the editor ! 

strudel_sounds_panel_update

